PR #2, opened 2025-06-23
Add option to use seeds for fragment extraction

sheridana:
The current fragment extraction using mutex computes segments per block (which are the supervoxels). It is sometimes desired to have more control over the fragmentation of the supervoxels (i.e when using a proofreading tool that can make use of the RAG for mincuts). A simple way to do this without needing seeded watershed is to just pass seeds to the mws function. Here the seeds are optionally computed on affinities as the local maxima of a distance transform.

### DIFF
--- a/volara/blockwise/extract_frags.py
+++ b/volara/blockwise/extract_frags.py
@@ -9,9 +9,11 @@ import numpy as np
 from funlib.geometry import Coordinate, Roi
 from funlib.persistence import Array
 from pydantic import Field
-from scipy.ndimage import gaussian_filter, measurements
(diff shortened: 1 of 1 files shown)


volara/blockwise/extract_frags.py
```diff
@@ -100,6 +102,11 @@ class ExtractFrags(BlockwiseTask):
     If using strides, you may want to switch from a grided stride to a random probability of
     filtering out an affinity. This can help avoid grid artifacts in the fragments.
     """
+    min_seed_distance: int | None = 10
```
pattonw, line 105: The default should probably be `None` just to match the doc string and maintain backwards compatibility and not change previously written pipelines.

Issue #11, opened 2025-07-14
Integrate Cloud Volume Interface


jakobtroidl:
* Interface with CloudVolume Datasets
* Currently used for sampling point clouds from segmentation volumes

jakobtroidl:
@pattonw @sheridana I want to contribute (1) my CloudVolume data interface for Volara and (2) my new custom blockwise task to sample point clouds from CloudVolume segmentation volumes. @sheridana suggested having the CloudVolume interface in volara (main repo) and the new custom blockwise task in volara-contrib. Does that work for you @sheridana @pattonw ?

jakobtroidl:
I'd be happy to make that change if we conclude here that it makes sense.

pattonw:
Yeah exactly. If you only need to use `CloudVolume` as a simple array, then I think they support getting a dask array. This can be wrapped with `funlib.persistence.Array` with the optional "voxel_size", "offset", "axis_names", "units", and "(axis)types" metadata. That would be fine.
If you need anything else from `CloudVolume`, like the `.mesh` or `.skeleton` features, then it won't fit well with the `Dataset` API and could just be its own class for now

jakobtroidl:
Yeah, focusing on the array functionality of CV seems to be fine for now. I will look into wrapping it with `funlib.persistence.Array`.

PR #12, opened 2025-07-15
feat: add cloud volume interface

jakobtroidl:
#11 Working on CloudVolume interface. @pattonw, are there any specific tests to include? This is still a work in progress.


volara/datasets.py
```diff
@@ -230,3 +232,37 @@ class Labels(Dataset):
     @property
     def attrs(self):
         return {}
+
+class CloudVolumeWrapper(Dataset):
+    """
+    Represents a volumetric dataset through Cloud Volume.
+    """
+    data_name: str
```
pattonw, line 240: do you need a `data_name` if you have a `store`? Can you parse the `store` url/path/str etc. to get a name?

volara/datasets.py
```diff
@@ -230,3 +232,37 @@ class Labels(Dataset):
...
+    dataset_type: Literal["volume"] = "volume"
+    mip: int = 0
+    timestamp: int = int(time.time()) # default to current time
+    voxel_size: PydanticCoordinate = Coordinate((1, 1, 1))
+    agglomerate: bool = True
+
+    @property
+    def data(self) -> CloudVolume:
```
pattonw, line 248: We want to use:
```
    def array(self, mode: str = "r") -> Array:
        return Array(cloud_volume_as_dask_array, ...)
```
to interface with the data.

volara/datasets.py
```diff
@@ -230,3 +232,37 @@ class Labels(Dataset):
...
+        vol = CloudVolume(self.store, mip=self.mip, use_https=True, agglomerate=self.agglomerate, timestamp=self.timestamp)
+        return vol
+
+    @property
+    def data_shape(self) -> list[int]:
```
pattonw, line 253: shouldn't need this if we can stick to the `array` interface

volara/datasets.py
```diff
@@ -230,3 +232,37 @@ class Labels(Dataset):
...
+        return self.data.shape
+
+    @property
+    def name(self) -> str:
+        return self.data_name
+
+    @property
+    def roi(self) -> Roi:
```
pattonw, line 261: also shouldn't need this if we stick to the `array` interface

volara/datasets.py
jakobtroidl:
```
arr = cv.to_dask()
arr = cloudvolume.dask.from_cloudvolume(cloudpath) # same as to_dask
```
@pattonw, I will use this syntax then.

jakobtroidl:
@pattonw, updated this PR, and I am now dynamically constructing the number of spatial dimensions, while always assuming the channel dimension is last. See the cloud-volume issue here: https://github.com/seung-lab/cloud-volume/issues/672. lmk if this looks good or if I should change something.

jakobtroidl:
Great. Thanks @pattonw.

PR #24, opened 2026-05-11
Add LambdaTask

rhoadesScholar:
Introduce generalized `LambdaTask` to apply a callable to a numpy input via Valor task distribution.


pattonw:
The main issue I see here is the `Callable` argument. I expect this to fail when you try to run the task on an async worker. The problem is that we need to serialize the task and send it to a worker. We generally do this by serializing each task to a yaml, and then starting a worker via `volara blockwise-worker -c /path/to/serialized/task.yaml`. So for this to work the callable probably needs to be serialized to a file via pickle, and then the task config needs to serialize/deserialize the path to the pickled lambda instead of the actual callable.

This should be able to be tested using the `LocalWorker()` as the worker config

volara/blockwise/lambda_task.py
```diff
@@ -0,0 +1,89 @@
+from contextlib import contextmanager
+from typing import Callable, Literal
+
```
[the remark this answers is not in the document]
rhoadesScholar: Fixed

tests/test_lambda_task.py
```diff
@@ -0,0 +1,107 @@
...
+def test_lambda_task_init_and_drop(zarr_2d, tmp_path):
+    """init() creates output zarr, drop_artifacts() removes it."""
+    raw_path, _ = zarr_2d
+    out_path = tmp_path / "test.zarr" / "out"
+    task = LambdaTask(
+        in_data=Raw(store=raw_path),
+        out_data=Labels(store=out_path),
+        lambda_func=lambda x: (x > 0.5).astype(np.uint8),
+        block_size=Coordinate(10, 10),
+    )
```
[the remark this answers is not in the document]
rhoadesScholar: @copilot apply changes based on [this feedback](https://github.com/e11bio/volara/pull/24#discussion_r3222335889)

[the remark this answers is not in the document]
rhoadesScholar: That did not go well 😂 Reverted Copilot's changes

rhoadesScholar:
> The main issue I see here is the `Callable` argument. I expect this to fail when you try to run the task on an async worker. The problem is that we need to serialize the task and send it to a worker. We generally do this by serializing each task to a yaml, and then starting a worker via `volara blockwise-worker -c /path/to/serialized/task.yaml`. So for this to work the callable probably needs to be serialized to a file via pickle, and then the task config needs to serialize/deserialize the path to the pickled lambda instead of the actual callable.
> 
> This should be able to be tested using the `LocalWorker()` as the worker config

@pattonw Added [a test using LocalWorker and two workers](https://github.com/e11bio/volara/pull/24/changes#diff-c687576c8b250169ca1a8786e7f6e66693392f2318372fd9ecc8438f833f2da3R88) Then added some Pydantic classes so it passes the test. Let me know what you think. :)

volara/utils.py
```diff
@@ -86,6 +86,46 @@ def validate_from_coordinates(data: tuple[Coordinate, Coordinate]) -> Roi:
...
+            if isinstance(value, str):
+                return from_b64(value)
+            raise ValueError(
+                f"Expected a callable or base64-encoded callable, got {type(value).__name__}"
+            )
+
+        return core_schema.json_or_python_schema(
+            json_schema=from_b64_schema,
+            python_schema=core_schema.no_info_plain_validator_function(callable_or_b64),
+            serialization=core_schema.plain_serializer_function_ser_schema(to_b64),
```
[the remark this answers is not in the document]
rhoadesScholar: The trust boundary is actually closed. The config is written and read by the same user, in the same invocation, on the same machine. The "unpickling untrusted data" warning applies when someone else controls the bytes.

volara/blockwise/lambda_task.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    lambda_func: PydanticCallable
+    """
+    The lambda function to apply to your dataset.
+    """
+    block_size: PydanticCoordinate
+    fit: Literal["shrink"] = "shrink"
+    read_write_conflict: Literal[False] = False
+    _out_array_dtype: np.dtype = np.dtype(np.uint8)
+
```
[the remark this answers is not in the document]
rhoadesScholar: Addressed in https://github.com/e11bio/volara/pull/24/commits/b35fefb65d157f82aa45d48019bb4133d650aaee

pattonw:
Oh nice! So it looks like you use pickle to dump the lambda to a binary string that gets written into / read from the yaml via the pydantic callable.
IDK what that yaml will look like. Probably not human readable, but I don't think we've really had anyone trying to manually create the configs and they mostly only get used for starting remote workers so I'm not too worried about them staying human read/write-able.

I like it

rhoadesScholar:
> ...
> 
> I like it

@pattonw Awesome! I figured we weren't too worried about someone reading the config files :) 
Mind adding an approving review -- or should I force merge?

examples/benchmarks/benchmark_cremi.py
```diff
@@ -15,15 +15,15 @@
 os.chdir(Path(__file__).parent)
 print(Path.cwd())
 
-raw = Raw(store="sample_A+_20160601.zarr/raw", scale_shift=(1 / 255, 0), writable=False)  # type: ignore[arg-type]
-affs = Affs(store="sample_A+_20160601.zarr/affs", writable=False)  # type: ignore[arg-type]
+raw = Raw(store="sample_A+_20160601.zarr/raw", scale_shift=(1 / 255, 0), writable=False)  # type: ignore
```
pattonw: interesting that the more specific type ignores started failing. Do you know why?

tests/test_lambda_task.py
```diff
@@ -0,0 +1,132 @@
...
+    task.init()
+
+    with task.process_block_func() as process_block:
+        process_block(block_2d)
+
+    result = task.out_data.array("r")[:]
+    expected = (data > 0.5).astype(np.uint8)
+    np.testing.assert_array_equal(result, expected)
```
pattonw, line 64: does this check dtype?

volara/blockwise/lambda_task.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    out_data: PydanticDataset
+    """
+    The output dataset after applying the lambda function.
+    """
+    lambda_func: PydanticCallable
+    """
+    The lambda function to apply to your dataset.
+    """
+    out_array_dtype: np.dtype | None = None
```
pattonw, line 30: I think this field exists by default on all blockwise tasks (https://github.com/e11bio/volara/blob/3b048c28bd21a62763daff3634c0ce6183771b0e/volara/blockwise/blockwise.py#L52). Do you need a custom one?

volara/blockwise/lambda_task.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    The dtype of the output array. If None, it will be the same as the input array.
+    """
+    block_size: PydanticCoordinate
```
pattonw, line 34: surprised this doesn't come by default on the `BlockwiseTask`...
Maybe an oversight on my part. It seems fundamental to a blockwise task.

Might be because sometimes the blocksize is determined by an argument (pytorch predict gets the blocksize from the model definition and doesn't make much sense to give people ability to override)

examples/benchmarks/benchmark_cremi.py
rhoadesScholar: right? no idea why :/ my bet is a version thing somewhere.

volara/blockwise/lambda_task.py
```diff
@@ -0,0 +1,96 @@
...
+    fit: str = "shrink"
+    read_write_conflict: bool = False
```
pattonw, line 36: These also come from base class by default. Probably don't need to override
https://github.com/e11bio/volara/blob/3b048c28bd21a62763daff3634c0ce6183771b0e/volara/blockwise/blockwise.py#L57

examples/benchmarks/benchmark_cremi.py
rhoadesScholar: 2b93258d3c5af4d40f3b46f56ec76d0c0fe904ee is the first commit with `ty` failing from this

volara/utils.py
```diff
@@ -86,6 +86,48 @@ def validate_from_coordinates(data: tuple[Coordinate, Coordinate]) -> Roi:
...
+        _source_type: Any,
+        _handler: Any,
+    ) -> core_schema.CoreSchema:
+        import base64
+
+        def from_b64(value: str) -> Callable:
+            import pickle
+
+            # Safe under the assumption that config files are user-owned (mode 700 tmp dirs).
+            # Do not use in contexts where config files may be written by untrusted parties.
```
pattonw, line 105: scary but we know pickle is a safety risk. Not sure if we need to care that much about this right now. I wonder if we could store a hash of the config somewhere to detect any hampering. Something to think about for future work if security becomes more of a concern

volara/utils.py
```diff
@@ -86,6 +86,48 @@ def validate_from_coordinates(data: tuple[Coordinate, Coordinate]) -> Roi:
...
+            return pickle.loads(base64.b64decode(value))
+
+        def to_b64(func: Callable) -> str:
+            import cloudpickle
```
pattonw: why not import at top?

pyproject.toml
```diff
@@ -31,6 +31,7 @@ dependencies = [
     "scikit-image>=0.25.2",
     "cloud-volume>=12.4.1",
     "polars>=1.31.0",
+    "cloudpickle>=3.0",
```
pattonw, line 34: why new dependency?

examples/benchmarks/benchmark_cremi.py
pattonw: A commit that has nothing to do with type checking...

If its just a versioning issue I'd rather figure that out and stick with the more specific ignores. But also not your responsibility if you just want to leave ty failing for now

volara/blockwise/lambda_task.py
rhoadesScholar: @pattonw Can you set it during init even if it's marked as private (_out... vs. out...)?

volara/blockwise/lambda_task.py
rhoadesScholar: Ah. I just copied over from the threshold task and then adjusted. I think you're right -- I'll remove them (fit too).

volara/blockwise/lambda_task.py
rhoadesScholar: I just copied over from the threshold task and then adjusted. Will remove.

volara/blockwise/lambda_task.py
rhoadesScholar: @pattonw It actually looks like `block_size` isn't in `BlockwiseTask`...

volara/blockwise/lambda_task.py
pattonw: Yeah the threshold task is one that could probably be removed after adding this one :)
Same with argmax.

A satisfying followup pull request would be to remove nodes that are sufficiently simple to cover with the LambdaTask, clean up the tests (or use LambdaTask to pass them if useful).
Could either remove the old tasks entirely, or make them subclasses of LambdaTask where the lambda is pre-defined for you. This would be more backwards compatible

tests/test_lambda_task.py
rhoadesScholar: it does now :D https://github.com/e11bio/volara/pull/24/commits/54727298629525a0ea724a997bea0152665bb0fa

volara/blockwise/lambda_task.py
rhoadesScholar: > make them subclasses of LambdaTask where the lambda is pre-defined for you.

I like this one.
It is worth noting that setting `fit` and `read_write_conflict` explicitly as defaults simplifies the call a lot. I could type them as `Literal` if you like. Note that, since the task doesn't have any way to define context, `read_write_conflict` is always False. Might want to add a way to define context though...

volara/blockwise/lambda_task.py
pattonw: That is a great question, and it doesn't look like it works as I wanted it to. It probably also doesn't belong here. This seems like it should belong to the `Dataset` that is set as the output, which should have its own dtype and tasks using it should respect that.

That is out of scope for this PR though so I think your current implementation is good.

volara/utils.py
rhoadesScholar: fixed 👍🏼

pyproject.toml
rhoadesScholar: regular pickle can only serialize functions that are importable by name — i.e., defined at the top level of a module that exists on disk. It serializes a reference to the function (module.funcname), not the function's actual bytecode.

cloudpickle serializes the function's bytecode, closure variables, and globals directly into the byte stream.

volara/blockwise/lambda_task.py
rhoadesScholar:
```
>       task = LambdaTask(
            in_data=Raw(store=raw_path),
            out_data=Labels(store=out_path),
            lambda_func=lambda x: (x > 0.5).astype(np.uint8),
            block_size=Coordinate(10, 10),
            _out_array_dtype=np.dtype(np.uint8),
        )
E       pydantic_core._pydantic_core.ValidationError: 1 validation error for LambdaTask
E       _out_array_dtype
E         Extra inputs are not permitted [type=extra_forbidden, input_value=dtype('uint8'), input_type=UInt8DType]
E           For further information visit https://errors.pydantic.dev/2.13/v/extra_forbidden
```

Can't pass kwargs to hidden variables (`_var_name`)

rhoadesScholar: For cases like `Relabel` I can understand hardcoding it, but not sure we want to do that for a lambda function. So it needs to be passable during init

volara/blockwise/lambda_task.py
rhoadesScholar: Hmmm. `Relabel` also strictly specifies `fit` and `read_write_conflict`: https://github.com/e11bio/volara/blob/1d8d63f9dd8c96acfb167df74df8605d9c6f74f0/volara/blockwise/relabel.py#L36-L37

volara/blockwise/lambda_task.py
rhoadesScholar: We could define a default in BlockwiseTask (e.g. `block_size = Coordinate((256,) * self.write_size.dims`)

pattonw: I think for now thats out of scope for this pull request

volara/blockwise/lambda_task.py
pattonw: Oh yeah, you're right. you're overriding the option with reasonable defaults for this task. All good

PR #26, opened 2026-06-02
Helpers for accessing different store endpoints

sheridana:
Adds a `prepared_store` property to Dataset which returns the actual store to pass to prepare_ds and open_ds. This way different s3 store endpoints can be read from / written to

```py
# local path to store
dataset = Dataset(
    store=Path("/path/to/data.zarr/dataset"),
)

# previous s3 store behavior
dataset = Dataset(
    store="s3://bucket/path/to/data.zarr/dataset",
)

# updating store endpoint url
dataset = Dataset(
    store="s3://bucket/path/to/data.zarr/dataset",
    s3_kwargs={
        "anon": False,
        "profile": "profile",
        "client_kwargs": {
            "endpoint_url": "endpoint_url",
            "region_name": "region_name",
        },
    },
)
```

### DIFF
--- a/volara/datasets.py
+++ b/volara/datasets.py
@@ -102,6 +102,7 @@ class Dataset(StrictBaseModel, ABC):
         """
         if not isinstance(self.store, Path):
             if isinstance(self.store, str) and self.store.startswith("s3://"):
+                # drop an s3 zarr
                 fs = self._s3fs()
                 try:
                     fs.rm(self.store, recursive=True)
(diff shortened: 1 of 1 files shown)


volara/datasets.py
```diff
@@ -54,28 +54,56 @@ class Dataset(StrictBaseModel, ABC):
...
+    def _s3fs(self):
+        import s3fs  # type: ignore[unresolved-import]
+
+        if self.s3_kwargs is None:
+            return s3fs.S3FileSystem()
+
+        return s3fs.S3FileSystem(**self.s3_kwargs)
+
     @property
-    def name(self) -> str:
```
pattonw, line 58: What happened to name? I think we need a string identifier to be available on datasets to make unique logs

sheridana: not sure why that docstring got dropped, added it back. relevant change should be the non path return:

`return self.store.rstrip("/").split("/")[-1]`

which ensures that an s3 string with an ending backslash still gets a non empty string name (e.g "s3://bucket/foo.zarr/" -> "foo.zarr" rather than "").

PR #32, opened 2026-07-08
run_blockwise returns the daisy TaskState map (surface failed/orphaned blocks)

rhoadesScholar:
## Problem
`BlockwiseTask.run_blockwise` returned an inconsistent type: `dict[str, TaskState]` on the serial path but a `bool` on the multiprocessing path (via `daisy.run_blockwise`). Worse, that bool reports `True` even when blocks fail, because `TaskState.is_done()` counts failed/orphaned blocks as "done" — so a partial output looks successful and no caller can tell.

## Change
Return daisy's `{task_id: TaskState}` map for **both** paths. The mp path now drives daisy's `Server` directly (via `_run_blockwise_with_states`, mirroring `daisy.convenience._run_blockwise` incl. the ThreadPool/stop_event so `KeyboardInterrupt` still stops the server cleanly) instead of the bool-collapsing convenience wrapper. Callers can now inspect `failed_count` / `orphaned_count` / `is_done()` and react to an incomplete run instead of silently accepting a partial output.

Downstream (slabreg) uses this to raise on any failed/orphaned block so a partial output never flows to the next stage.

https://claude.ai/code/session_019H8XdbRTqyaVJPdHmjYre5


rhoadesScholar:
Needs tests

pattonw:
This seems like something that should be fixed in `daisy`, not worked around in `volara`

rhoadesScholar:
Added tests in `607f052` (`tests/test_run_blockwise_states.py`, 4 cases): a serial `BlockwiseTask` returns the `{task_id: TaskState}` dict with the right `completed_count`; a multiprocessing run whose worker raises on a block returns `failed_count > 0` **while `is_done()` is still `True`** — the exact failure the old collapsed bool hid; and a `Pipeline` (sequential `+`) returns the merged map on both the serial and multiprocessing paths.

Re: fixing it in daisy — agreed, and that's what daisy 2.0 does (the convenience wrapper stops collapsing the multiprocessing path to `all(is_done())`). Since 2.0 isn't merged to main yet, this PR is the bridge on 1.2.2. Two things worth separating:

- daisy already tracks and returns the states — `Server.run_blockwise` and `SerialServer.run_blockwise` both return `{task_id: TaskState}`. The only 1.2.2 collapse is `daisy.convenience._run_blockwise` doing `return all(task_state.is_done() for ...)`, and `TaskState.is_done()` counts failed/orphaned blocks as done, so a fully-failed run reads `True`.
- Propagating that map through `BlockwiseTask.run_blockwise` and `Pipeline.run_blockwise` is our own API surface regardless of daisy — both were discarding what daisy handed back (a bool on the MP path, `None` for `Pipeline`).

Plan: land this propagation now; once we're on daisy 2.0 we delete the duplicated `_run_blockwise_with_states` helper (a verbatim mirror of daisy's ThreadPool/`stop_event` boilerplate) and call daisy directly. The helper is deliberately isolated so that swap is a one-liner, and the tests pin the return contract, not the helper.

volara/blockwise/blockwise.py
```diff
@@ -27,33 +26,17 @@
...
-def _run_blockwise_with_states(tasks):
-    """``daisy.run_blockwise``'s multiprocessing path, but returning daisy's
-    ``{task_id: TaskState}`` map instead of collapsing it to a bool (which hides failed /
-    orphaned blocks). Mirrors ``daisy.convenience._run_blockwise`` and its ThreadPool /
-    stop_event handling so a ``KeyboardInterrupt`` still stops the server cleanly. Defined at
-    module scope because ``BlockwiseTask.run_blockwise``'s ``multiprocessing`` parameter
-    shadows the stdlib module inside the method."""
-    from multiprocessing import Event
-    from multiprocessing.pool import ThreadPool
+def _to_daisy_roi(roi: Roi) -> daisy.Roi:
```
rhoadesScholar: @pattonw Seems like this shouldn't be necessary. Will daisy handle a funlib Roi being passed to daisy.Block() and daisy.Task()?

volara/blockwise/blockwise.py
```diff
@@ -264,7 +250,14 @@ def worker_func(self):
 
             def run_worker():
```
rhoadesScholar, line 243: @pattonw Thoughts on this, now that daisy v2 wants workers to be blocking?

volara/blockwise/blockwise.py
```diff
@@ -26,6 +25,30 @@
...
+# daisy v2 removed the "no timeout" option: ``Task(timeout=None)`` now means the
+# built-in default (~600s) rather than 1.x's "never time out", and non-positive
+# values raise ``ValueError`` (see daisy's MIGRATION.md). Blocks that legitimately
+# run longer than the default -- e.g. in-process GPU descriptor / SOFIMA-flow steps
+# -- would be reclaimed and abandoned mid-run. To preserve 1.x behaviour we map an
+# unset ``block_timeout`` to this large-but-finite sentinel (24h), which never trips
+# on a real block yet still lets daisy eventually reclaim a genuinely hung worker
+# (strictly safer than 1.x, which waited forever). Override per task via
+# ``block_timeout`` when reclaim-on-hang is wanted sooner.
+_V1_COMPAT_NO_TIMEOUT = 24.0 * 60.0 * 60.0
```
pattonw: no, we should not have blocks processing for more than 10 minutes generally. We don't want to let blocks hang for 24 hours. If thats is happening regularly we need to re-visit a task.
The only exception I can think of at the moment is the global graph mutex watershed which should take a long time on large datasets because its a global operation. For that one specifically we will want to override the time-out to maybe 24 hours.

volara/blockwise/blockwise.py
pattonw: No this should not be necessary. Why was it added?

volara/blockwise/blockwise.py
```diff
@@ -63,6 +86,15 @@ class BlockwiseTask(StrictBaseModel, ABC):
     Whether blocks have read/write dependencies on neighborhing blocks requiring
     a specific ordering to the block processing to compute a seamless result.
     """
+    block_timeout: float | None = None
```
pattonw, line 65: it should not be allowed to be None. All blocks should time out eventually. I'd set the default the same as daisy v2 at 10 minutes. the whole point of processing blocks is they should be small and process quickly. generally in the order of a few seconds up to maybe a minute.
Major exception mentioned above already

volara/blockwise/blockwise.py
```diff
@@ -162,8 +194,11 @@ def process_roi(self, roi: Roi, context: Coordinate | None = None):
         A helper function to process a given roi without needing to start a
         whole blockwise job.
         """
+        read_roi = roi if context is None else roi.grow(context, context)
+        # daisy.Block is the raw Rust ctor (NOT shimmed by v1_compat): funlib Rois
+        # must be converted to native daisy.Rois here.
         block = daisy.Block(
```
pattonw: I don't think this is necessary? Shouldn't be at least. `daisy.Block` should be the same as v1 `daisy.Block` but maybe I didn't test that. Definitely a bug that should be fixed asap upstream

volara/workers.py
```diff
@@ -141,6 +211,11 @@ def get_slurm_command(
         # Initialize the command list with the base sbatch command
         run_command = ["sbatch"]
```
pattonw: We use "sbatch" which is non-blocking.
We should probably just use "srun" which is blocking.
No need to have a custom wrapper around subprocess.run() with polling by process name etc.

pattonw:
The title/description of this pull request is not related to the actual pull request anymore. This should be closed, re-submitted with a better title and description of its purpose

rhoadesScholar:
Closing and resubmitting, per @pattonw's review: *"The title/description of this pull request is not
related to the actual pull request anymore. This should be closed, re-submitted with a better title
and description of its purpose."*

You were right on both counts, and on a third I hadn't noticed: the description said the
multiprocessing path goes *"via `_run_blockwise_with_states`, mirroring `daisy.convenience._run_blockwise`"*,
but commit c9e10a0 on this same branch **deleted that helper** in favour of daisy v2's native
`Server.run_blockwise`. So the code being described no longer existed. My "bridge on 1.2.2" reply was
stale for the same reason — 10f6a87 pins daisy v2 on this branch.

Auditing the diff, this PR was **six changes in one**: the TaskState-map feature, a daisy-v2
dependency change, the funlib→daisy Roi/Block boundary removal, the worker-model refactor, a
`graph_mws` timeout change, a `dbs` fix, and a ruff pass. History also carries duplicated commit pairs
(`fa4e97b`/`b412d99`, `c5b5483`/`fbb47f1`, `48c89ff`/`bac0459`) from an earlier merge.

## Resubmitted, independent of daisy v2 — all three target `main` on daisy 1.x

| | |
|---|---|
| #41 | `feat(blockwise): return the daisy TaskState map from run_blockwise on both paths` — the actual purpose of this PR. **233/-8 across 3 files** instead of 529/-118 across 11, with a runnable failing→passing reproduction in the description. |
| #42 | `deps(daisy): cap daisy<2 until the v2 migration lands` — one line. This was step 1 of #34 and never landed because this branch went straight to the git `v2.0` pin. Details below; it matters more than it looks. |
| #43 | `fix(dbs): stop drop()/drop_edges() swallowing real Postgres failures` |

## Still to come, and deliberately not reopened yet

The daisy-v2 half needs daisy 2.0.0 to be installable from PyPI first. It is not: `funkelab/daisy`'s
`publish.yaml` gates `release` on a `test-wheel` job that currently fails, so tagging 2.0.0 today
builds every wheel and then does not publish. Merging a `daisy @ git+…@v2.0` direct URL here would
also break volara's own `publish.yaml` — PyPI rejects PEP 440 direct references in `Requires-Dist`,
and `[tool.uv.sources]` is uv-only metadata that never reaches the built distribution.

So, once daisy 2.0.0 is real, as a stack rather than four PRs against `main`:

1. `refactor(daisy-v2): migrate to daisy v2` — and **first re-check whether it's still needed.** Your
   review said the Roi/Block conversion layer *"should not be necessary… `daisy.Block` should be the
   same as v1 `daisy.Block`"*. daisy has since landed `585cc1a` (compat `Block` accepts and presents
   funlib ROIs) and `26f10ed` (unify compat blocks), which may remove that layer entirely.
2. `refactor(workers): blocking srun / bsub -K per daisy v2's spawn contract` — **with #40 squashed
   in**, see below.
3. `fix(graph_mws): override block_timeout to 24h for the global agglomeration block` — done your way:
   keep daisy v2's 600 s default rather than allowing `None`, and override only the global
   mutex-watershed task. You were right that `None` is the wrong default, and daisy v2 now enforces
   600 s anyway, so `None` is no longer even expressible.

## On #40, and why it will not be a separate PR

#40 removes 28 duplicated lines from `get_slurm_command`. Checking where the duplication came from:
**it does not exist on `main`.** `main` builds `sbatch --wrap=<command>` with no duplication at all —
the doubling was introduced by *this branch's own* `srun` refactor (`b412d99`, itself duplicated as
`fa4e97b`).

So #40 is not a bug fix against `main`; it is a fixup to unmerged work. Rather than land a "fix" for a
bug `main` never had, `c3cf034` gets squashed into `b412d99` in item 2 above, so the duplication never
exists in the history. #40 will be closed pointing here.

For the record, since the doubled command cost real full-volume runs — verified by constructing the
command both ways:

```
base    srun … --gpus=1 --gpus=1 … --output=X --error=Y --output=X --error=Y \
             volara-cli blockwise-worker -c cfg volara-cli blockwise-worker -c cfg     (19 tokens)
fixed   srun … --gpus=1 … --output=X --error=Y volara-cli blockwise-worker -c cfg      (12 tokens)
```

Only the doubled worker command is harmful — click rejects it with *"Got unexpected extra arguments"*
and every block is orphaned. The doubled `--output`/`--error` carry identical values and the
`--job-name` is re-appended after the reassignment, so those are cosmetic; #40's description overstated
them and that will be corrected.

`tests/test_workers.py` **passes against the duplicated command** (3 passed either way) because it
asserts `cmd[-4:] == [...]` — the duplicate is appended, so the tail is still correct. Item 2 will add
`cmd.count("blockwise-worker") == 1`. slabreg already added exactly that downstream.

## On #39

The log-basedir break is a **daisy v2 regression**, which answers your question (1). daisy 1.x put
`logdir` into the worker `Context`, so it crossed any process boundary via `DAISY_CONTEXT`:

```
daisy 1.x   Context(...).to_env() -> hostname=…:port=…:task_id=…:worker_id=…:logdir=<driver's>
daisy 2.0   Context(...).to_env() -> hostname=…:task_id=…:port=…:worker_id=…      # no logdir
```

v2's `Client.__init__` re-injects it from the process-global `get_log_basedir()`, which its own comment
describes as what *"fork-inheriting workers see as the value the master set"* — but volara's workers are
**spawned `volara-cli` processes, not forks**, so the global is daisy's default. No `KeyError`, no
crash, silently the wrong basedir. You were right that it worked before; the mechanism that made it
work was removed upstream.

So the primary fix belongs in daisy, as you suggested — I'll open that issue. #39's volara-side env-var
change then becomes a guard rather than the fix, and will be reframed and restacked onto item 1.

The branch stays; nothing here is lost.

PR #33, opened 2026-07-19
fix(workers): reap slurm worker jobs on teardown (stop the leak)

rhoadesScholar:
## The bug (intrinsic, all slurm users)

`SlurmWorker.get_command` builds a plain `sbatch … --wrap=<worker>` (no `--wait`) and
`BlockwiseTask.worker_func`'s `run_worker` does `subprocess.run(cmd)`. So `sbatch` **returns
the instant the job is queued**, the **job id is discarded**, and daisy's teardown
(`Worker.stop()` → `process.terminate()`) only kills the already‑exited local `sbatch`
client — it can **never `scancel` the actual slurm job**. There is no `scancel` anywhere.

Any worker that doesn't cleanly self‑exit on `block is None` (e.g. a worker that finishes
connecting *after* its `run_blockwise` server has shut down) therefore **leaks until
walltime**. Across a driver that runs many `run_blockwise` calls (one per stage), these pile
up: we observed a 3‑slab run leak **96 idle GPU workers** from the sequential lowres‑predict
stages, saturating the queue so the next stage's workers got no blocks and the run stalled
for ~8 h on ~99 idle nodes.

## The fix

- **`Worker.run(cmd)`** — new hook that runs one worker to completion and *blocks for its
  lifetime* (daisy relies on this: it terminates the process running `run` to stop the
  worker). Base/local = `subprocess.run` (a child process daisy can terminate directly).
- **`SlurmWorker.run`** — submit with `--parsable` to capture the job id, poll `squeue` until
  the job finishes, and `scancel` it in a `finally` **and** a `SIGTERM` handler → the job is
  reaped on normal completion, on daisy teardown, and on driver abort. `sbatch` now also
  `--job-name`s the worker after its task.
- **`blockwise.run_worker`** — call `worker_config.run(cmd)` instead of `subprocess.run(cmd)`.

This also fixes a latent correctness issue: fire‑and‑forget made `run_worker` return
instantly, so daisy saw the worker "finish" immediately; `run` now blocks for the real job
lifetime, matching the local backend's contract.

## Tests

`tests/test_slurm_worker_teardown.py` simulates `sbatch`/`squeue`/`scancel` (no cluster):
reaps on normal completion, raises on submit failure, and reaps on `SIGTERM` (teardown).

Stacked on #32 (base `feat/run-blockwise-states`). LSF has the same latent pattern; left as a
follow‑up (untestable here).

### DIFF
--- a/volara/blockwise/blockwise.py
+++ b/volara/blockwise/blockwise.py
@@ -538,15 +538,25 @@ class BlockwiseTask(StrictBaseModel, ABC):
         """
         with self.task(multiprocessing=multiprocessing) as task:
             tasks = [task]
-            if multiprocessing:
-                # daisy v2's Server.run_blockwise returns the {task_id: TaskState}
-                # map natively, so the old 1.x ThreadPool/IOLooper/progress-monitor
-                # states workaround is no longer needed.
-                return daisy.Server().run_blockwise(tasks)
-            # Serial path: module-level run_blockwise returns a bool unless
-            # return_states=True (see daisy._runner), so request the states map to
-            # keep the same {task_id: TaskState} contract as the distributed path.
-            return daisy.run_blockwise(tasks, multiprocessing=False, return_states=True)
+            try:
+                if multiprocessing:
+                    # daisy v2's Server.run_blockwise returns the {task_id: TaskState}
+                    # map natively, so the old 1.x ThreadPool/IOLooper/progress-monitor
+                    # states workaround is no longer needed.
+                    return daisy.Server().run_blockwise(tasks)
+                # Serial path: module-level run_blockwise returns a bool unless
+                # return_states=True (see daisy._runner), so request the states map to
+                # keep the same {task_id: TaskState} contract as the distributed path.
+                return daisy.run_blockwise(
+                    tasks, multiprocessing=False, return_states=True
+                )
+            finally:
+                # Teardown sweep: reap any cluster worker jobs that outlived the run. daisy has
+                # already stopped its pools and every block is terminal here, so a name-scoped
+                # scancel cannot touch in-flight work. Workers are named after the task, so the
+                # sweep is scoped to this task's jobs. No-op for local workers. See Worker.cleanup.
+                if self.worker_config is not None:
+                    self.worker_config.cleanup(self.task_name)
 
     def __add__(self, other: "BlockwiseTask | Pipeline") -> "Pipeline":
         """
--- a/volara/blockwise/pipeline.py
+++ b/volara/blockwise/pipeline.py
@@ -167,17 +167,26 @@ class Pipeline:
 
             all_tasks = list(task_map.values())
 
-            if multiprocessing:
-                # daisy v2's Server.run_blockwise returns the {task_id: TaskState}
-                # map natively (honouring the v1_compat upstream_tasks edges each
-                # task carries), so a Pipeline surfaces failed/orphaned blocks too.
-                return daisy.Server().run_blockwise(all_tasks)
-            # Serial path: module-level run_blockwise returns a bool unless
-            # return_states=True (see daisy._runner); request the states map to
-            # match the distributed path's contract.
-            return daisy.run_blockwise(
-                all_tasks, multiprocessing=False, return_states=True
-            )
+            try:
+                if multiprocessing:
+                    # daisy v2's Server.run_blockwise returns the {task_id: TaskState}
+                    # map natively (honouring the v1_compat upstream_tasks edges each
+                    # task carries), so a Pipeline surfaces failed/orphaned blocks too.
+                    return daisy.Server().run_blockwise(all_tasks)
+                # Serial path: module-level run_blockwise returns a bool unless
+                # return_states=True (see daisy._runner); request the states map to
+                # match the distributed path's contract.
+                return daisy.run_blockwise(
+                    all_tasks, multiprocessing=False, return_states=True
+                )
+            finally:
+                # Teardown sweep for EVERY task in the pipeline: reap cluster worker jobs that
+                # outlived the run. daisy has stopped its pools and all blocks are terminal here,
+                # so the name-scoped scancel per node cannot touch in-flight work. No-op for
+                # local workers. See Worker.cleanup.
+                for node in node_ordering:
+                    if node.worker_config is not None:
+                        node.worker_config.cleanup(node.task_name)
 
     def drop(self):
         for task in self.task_graph.nodes():
--- a/volara/dbs.py
+++ b/volara/dbs.py
@@ -255,17 +255,25 @@ class PostgreSQL(DB):
     def drop(self) -> None:
         try:
             db = self.open("r+")
-            db._drop_tables()
-            db._create_tables()
-        except RuntimeError:
-            # DB doesn't exist yet
-            pass
+        except RuntimeError as e:
+            # A fresh database has no schema/metadata yet, so open('r+') (a read mode)
+            # raises "metadata does not exist". There is nothing to drop, so this is a
+            # successful no-op -- drop() must NEVER bring a database into existence
+            # (DB.init() is the only creator). Re-raise anything else so real failures
+            # (e.g. a broken connection) aren't silently swallowed.
+            if "metadata does not exist" in str(e):
+                return
+            raise
+        db._drop_tables()
+        db._create_tables()
 
     def drop_edges(self) -> None:
         try:
             db = self.open("r+")
-            db._drop_edges()
-            db._create_tables()
-        except RuntimeError:
-            # DB doesn't exist yet
-            pass
+        except RuntimeError as e:
+            # Fresh DB (no schema): nothing to drop, no-op. See drop() -- must not create.
+            if "metadata does not exist" in str(e):
+                return
+            raise
+        db._drop_edges()
+        db._create_tables()
--- a/volara/workers.py
+++ b/volara/workers.py
@@ -34,6 +34,17 @@ class Worker(StrictBaseModel, ABC):
         both this call and the driver (see SlurmWorker.run)."""
         sp.run(cmd)
 
+    def cleanup(self, task_name: str) -> None:
+        """Reap any workers for ``task_name`` that outlived the run. Called when a task's
+        ``run_blockwise`` returns (after daisy has stopped its worker pools and every block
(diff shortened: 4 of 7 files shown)


pattonw:
Tests seem to be failing due to sbatch not being available on the github action node. Tests should either:
1) spoof sbatch, and not need it, run anywhere every time
2) actually test sbatch, run on a flag, e.g. `pytest --test-sbatch ...` only when we explicitly want to test sbatch

volara/blockwise/blockwise.py
```diff
@@ -264,7 +264,10 @@ def worker_func(self):
 
             def run_worker():
                 cmd = worker_config.get_command(config_file, self.task_name)
-                return subprocess.run(cmd)
+                # worker_config.run blocks for the worker's lifetime and (for cluster backends)
+                # reaps the submitted job on completion/teardown. A plain subprocess.run(cmd)
+                # here fire-and-forgot slurm jobs and leaked them (see SlurmWorker.run).
+                return worker_config.run(cmd)
```
pattonw: It makes me nervous that the `worker.run` command now blocks for the duration of the worker.
That may be a better path to go but I think would require careful consideration. I don't think `daisy` expects worker functions to block on call. What happens if there's a bug in a worker and it hangs and never unblocks?

volara/blockwise/blockwise.py
```diff
@@ -508,11 +511,17 @@ def run_blockwise(
...
+                if multiprocessing:
+                    return _run_blockwise_with_states(tasks)
+                server = daisy.SerialServer()
+                _cl_monitor = CLMonitor(server)  # noqa: F841
+                return server.run_blockwise(tasks)
+            finally:
+                # Reap any workers that outlived the run (daisy respawn / orphaned launcher).
+                # Workers are named after the task, so scancel-by-name is a safe teardown sweep.
+                if self.worker_config is not None:
+                    self.worker_config.cleanup(self.task_name)
```
pattonw, line 559: A worker cleanup seems like a good idea. Especially for something like slurm where we know the names of the jobs we start and can just kill outstanding jobs when the task is done.

I only worry that it could hide other problems. If we have workers that aren't properly exiting, we should figure out why, not implement a brute force solution that may leave datasets or dbs in corrupted states.

volara/blockwise/pipeline.py
```diff
@@ -171,18 +171,26 @@ def run_blockwise(self, multiprocessing: bool = True):
...
+                    server = daisy.SerialServer()
+                    _cl_monitor = daisy.cl_monitor.CLMonitor(server)  # type: ignore[unresolved-attribute]
+                    return server.run_blockwise(all_tasks)
+            finally:
+                # Reap any workers that outlived the run for EVERY task in the pipeline (daisy
+                # respawn / orphaned launcher). Workers are named after their task, so a
+                # scancel-by-name sweep per node is a safe teardown. See Worker.cleanup.
+                for node in node_ordering:
+                    if node.worker_config is not None:
+                        node.worker_config.cleanup(node.task_name)
```
pattonw, line 189: This seems like a workaround for strange behavior from daisy. We should make this more consistent in `daisy` itself rather than work around it in `volara`

volara/dbs.py
```diff
@@ -258,14 +258,18 @@ def drop(self) -> None:
...
             db._create_tables()
         except RuntimeError:
-            # DB doesn't exist yet
-            pass
+            # Fresh database with no schema/metadata yet: 'r+' (a read mode) raises
+            # "metadata does not exist". Create it in 'w' mode, which drops+creates the tables
+            # AND stores the metadata, so a subsequent open('r+') (e.g. by the keep-main RAG
+            # tasks) succeeds. SQLite creates its tables on open; Postgres 'r+' does not, so a
+            # fresh PG database needs this or every consumer fails on first open.
+            self.open("w")
```
pattonw: database should not be created if we drop it

volara/dbs.py
```diff
@@ -258,14 +258,18 @@ def drop(self) -> None:
...
     def drop_edges(self) -> None:
         try:
             db = self.open("r+")
             db._drop_edges()
             db._create_tables()
         except RuntimeError:
-            # DB doesn't exist yet
-            pass
+            # fresh DB (no schema): create it (see drop() -- 'w' creates tables + metadata)
+            self.open("w")
```
pattonw: again database should not be created if we drop it

volara/workers.py
```diff
@@ -24,11 +24,29 @@ def get_command(self, config_path: Path, task_name: str) -> list[str]:
         ]
         return cmd
 
+    def run(self, cmd: list[str]) -> None:
+        """Run ONE worker to completion. MUST block for the worker's lifetime: daisy relies
+        on this (it terminates the process running ``run`` to stop the worker). The base /
```
pattonw: Does daisy rely on this? I thought it expected workers to just start and then communicated stop signals via tcp

volara/workers.py
```diff
@@ -24,11 +24,29 @@ def get_command(self, config_path: Path, task_name: str) -> list[str]:
...
+        local backend is a plain child process, which ``subprocess.run`` blocks on and daisy
+        can terminate directly. Cluster backends (slurm/lsf) submit an INDEPENDENT job that
+        is NOT a child of this process, so they MUST override ``run`` to babysit and reap it
+        -- otherwise the job outlives both this call and the driver (see SlurmWorker.run)."""
+        sp.run(cmd)
+
+    def cleanup(self, task_name: str) -> None:
+        """Reap any workers for ``task_name`` that outlived the run. Called when a task's
```
pattonw, line 38: Same concerns as above.
We don't know where/how the user will start workers (lsf, slurm, etc. etc.) so it makes sense that we should give users a custom way of reaping workers. If useful it should probably go in `daisy` as a `reap_workers` function attribute of the `Task`.
But I worry that overly relying on reaping workers will result in other problems getting hidden. The workers should cleanly exit on their own after no longer receiving blocks.

rhoadesScholar:
CI fixed in `338285d`: `tests/test_slurm_worker_teardown.py` now spoofs `sbatch`/`scancel`/`squeue` as PATH-prepended shim executables (each records its argv so we can assert the reap actually issued `scancel --name <task>`). The reap path runs for real, so it runs anywhere with no cluster. Added a `slurm` pytest marker with `addopts = "-m 'not slurm'"`, so any genuinely-slurm-requiring test is opt-in (`pytest -m slurm`) — after spoofing, none currently need it. `test_slurm_worker_teardown.py` + `test_dbs.py`: 7 passed, 2 skipped (PostgreSQL-parametrized, no PG server here).

volara/blockwise/blockwise.py
rhoadesScholar: You're right — reverted in `338285d`. `run_worker()` is back to a plain `subprocess.run(cmd)`; I dropped `SlurmWorker.run`/`_job_active`/`poll_interval` and the `--parsable` id-capture. Blocking daisy's `process_function` isn't needed for the leak fix: daisy runs the spawn function in its own `multiprocessing.Process`, and its clean stop is TCP-based (`acquire_block()` returns `None` → the worker self-exits). Reaping is handled entirely by the name-scoped teardown sweep, which avoids the hang case you flagged.

volara/blockwise/blockwise.py
rhoadesScholar: Kept, and I've narrowed the PR to just this. On the "could hide corruption" worry: the sweep runs in `run_blockwise`'s `finally`, i.e. after daisy's `server.run_blockwise()` has already returned. daisy's event loop only exits once every task `is_done()` (all blocks terminal), and its own `finally` calls `worker_pools.stop()` (terminate on every worker) before returning — so when our sweep fires, nothing is in flight and daisy has already tried to stop its workers; a name-scoped `scancel` can't interrupt a write. It also doesn't paper over the real bug: the worker's normal self-exit on `block is None` is unchanged. The sweep only catches the one case we can explain — a slurm job whose `sbatch` client exited at submit time, so daisy's `terminate()` reaps a dead client and never the real job. That's the ~99-node storm operators were clearing by hand with `scancel --name`; this just automates it, scoped to the task's own job name.

volara/blockwise/pipeline.py
rhoadesScholar: Agreed the consistent home is daisy — and daisy 2.0 already stops collapsing the multiprocessing path, so this becomes moot once we're on 2.0 (not yet merged to main). Note this helper (`_run_blockwise_with_states`) comes from the stacked TaskState-map PR (#32), not the teardown work here — #33 only wraps the existing call in `try/finally` for the reap sweep. It's a thin mirror of `daisy.convenience._run_blockwise` (same ThreadPool + `stop_event`); we drop it and call daisy directly once we move to daisy 2.0.

volara/dbs.py
rhoadesScholar: Fixed in `43c8d86`: the `except` branch no longer calls `open(..., "w")` — `drop` no-ops when the metadata doesn't exist and re-raises anything else, so dropping never creates the database.

volara/dbs.py
rhoadesScholar: Same fix as above (`43c8d86`), applied to `drop_edges`; added `test_drop_does_not_create` to pin it.

volara/workers.py
rhoadesScholar: Moot now — I removed `Worker.run` in `338285d`. To answer directly: yes, the clean stop is over TCP (the worker self-exits when `acquire_block()` returns `None`). daisy also force-stops a worker with `process.terminate()` + `join()` (`daisy/worker.py`), and its own in-process worker blocks for the worker's lifetime — so blocking wasn't *wrong*, but it wasn't needed, and I dropped it. daisy does not rely on a blocking `run()` for correctness.

volara/workers.py
rhoadesScholar: Fully agree the long-term home is daisy — a `reap_workers` hook on `Task` (or the worker backend) that daisy calls on teardown, so each scheduler supplies its own reaping. This is a scoped volara-side interim: name-scoped (`scancel --name <task>`, never a blanket cancel), fires only at `run_blockwise` teardown after daisy has stopped its pools and all blocks are terminal, and a no-op for local workers. We're not leaning on it to hide unhealthy workers — the worker still self-exits on `block is None`; the sweep only catches the one structural case daisy's `terminate()` can't reap (an already-exited `sbatch` client). Happy to move it to a daisy `reap_workers` hook once that lands and have volara consume it.

PR #35, opened 2026-07-27
Db bottleneck

sheridana:
This PR adds bulk write option to fragment extraction and affinity agglomeration. This allows for better db write scaling by rebuilding indexes at the end. It also adds a faster relabeling for larger segmentation volumes, which may be better in a separate PR. Both changes tested on large data and seem to scale well linearly.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -109,3 +109,154 @@ def block_2d() -> daisy.Block:
         read_roi=Roi((0, 0), (10, 10)),
         write_roi=Roi((0, 0), (10, 10)),
     )
+
+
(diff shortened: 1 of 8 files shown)


pattonw:
Does this depend on a specific version of funlib.persistence? It would probably be good to update the dependencies to make sure that whatever version we depend on supports this new behavior

sheridana:
oh hmm, I assumed just most recent changes where the bulk writes were introduced, but now I see more recent updates that affect the same files (e.g updating nodes in place). I can check with most recent version and add tests, then if all good will pin the funlib persistence version (might want to push a new persistence release first if so).

PR #37, opened 2026-07-27
Add fragment extraction optimizations

sheridana:
Adds optimizations to fragment extraction via:

 - noise generated directly into `shift`, affinities folded in in place, and the redundant `.astype(np.float64)` copy dropped - the original allocated up to six full (C, Z, Y, X) float64 volumes per block, this holds two
- per-fragment mean affinity and centroids via `np.bincount` instead of `scipy.ndimage.mean` / `center_of_mass`, which rescan the whole volume per label

- minor fix returns replace_values(..) if filtering fragments - was a silent no-op before. 
- minor fix computes ndim of affs so we dont hardcode to first 3 channels when taking mean for fragment filtering

Currently this runs via an "optimized" flag and uses a module level random generator for testing (reproducible noise etc). If the changes look reasonable, then I will remove the optimized flag to make them default and add appropriate tests.

### DIFF
--- a/volara/blockwise/extract_frags.py
+++ b/volara/blockwise/extract_frags.py
@@ -21,6 +21,12 @@ from .blockwise import BlockwiseTask
 
 logger = logging.getLogger(__file__)
 
+# Module-level generator so the affinity noise can be made reproducible for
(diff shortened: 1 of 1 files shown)


pattonw:
This is in the aff-agglom pull request, and already reviewed

PR #38, opened 2026-07-27
Optimize aff agglom

sheridana:
Adds optimizations to aff agglom via:

- relabel via `np.unique(return_inverse=True)` instead of a full-array scan per fragment (O(n_frags * n_voxels))
- affinity counting over only the boundary voxels, with an integer cantor pairing and `np.bincount`, instead of full-volume float64 pairings handed to `scipy.ndimage.mean` / `sum_labels`

Currently this runs via an "optimized" flag. If the changes look reasonable, then I will remove the optimized flag to make them default and add appropriate tests.

### DIFF
--- a/volara/blockwise/aff_agglom.py
+++ b/volara/blockwise/aff_agglom.py
@@ -5,7 +5,6 @@ from typing import Annotated, Callable, Generator, Literal
 
 import networkx as nx
 import numpy as np
-import scipy.ndimage
 from daisy import Block
 from funlib.geometry import Coordinate, Roi
 from funlib.math import inv_cantor_number
@@ -76,21 +75,6 @@ class AffAgglom(BlockwiseTask):
     We don't have read write conflicts in this task and can compute
     every block independently in an arbitrary order.
     """
-    optimized: bool = False
-    """
-    Use the optimized implementations of the fragment relabel and the affinity
-    counting. Produces the same edges and weights as the original code, just
-    faster (measured 7.6x on the relabel, 40x on the counting):
-
-    - relabel via `np.unique(return_inverse=True)` instead of a full-array scan
-      per fragment (O(n_frags * n_voxels))
-    - affinity counting over only the boundary voxels, with an integer cantor
-      pairing and `np.bincount`, instead of full-volume float64 pairings handed
-      to `scipy.ndimage.mean` / `sum_labels`
-
-    Defaults to False so the original behavior is what you get unless you ask
-    for it. Set True to compare.
-    """
 
     @property
     def task_name(self) -> str:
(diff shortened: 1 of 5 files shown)


volara/blockwise/aff_agglom.py
```diff
@@ -107,77 +122,121 @@ def agglomerate(self, affs: np.ndarray, frags: np.ndarray, rag: nx.Graph):
...
-            rev_mapping = {v: k for k, v in frag_mapping.items()}
-            for old, seq in frag_mapping.items():
-                frags[frags == old] = seq
-
-            if len(fragment_ids) == 0:
+            # Both branches relabel fragments to a dense 1..k range so the cantor
+            # pairings in count_affs stay small, and produce the same mapping.
+            if self.optimized:
+                # np.unique(return_inverse=True) IS this relabel, done in C.
+                unique_vals, inverse = np.unique(frags, return_inverse=True)
```
pattonw: This first chunk is very nice. We can go ahead and drop the optimized flag here and just go with the new version. This makes a lot of sense and avoids iterating over the array to do a masked reassignment.
The 0 handling seems a little awkward to me here, but fine for now

volara/blockwise/aff_agglom.py
```diff
@@ -107,77 +122,121 @@ def agglomerate(self, affs: np.ndarray, frags: np.ndarray, rag: nx.Graph):
...
+
+            mask = (offset_frags != base_frags) & (offset_frags > 0) & (base_frags > 0)
+
+            # Only voxels on a boundary between two distinct fragments carry an
+            # edge -- typically a few percent of the block. Select them first and
+            # do all the arithmetic on that small array, instead of computing
+            # full-volume cantor pairings and handing them to
+            # scipy.ndimage.mean/sum_labels (which then rescan every voxel; that
+            # pair alone was ~85% of this function).
+            u_frags = offset_frags[mask].astype(np.uint64)
```
pattonw, line 163: This is great, makes a lot of sense. I thought the bulk of the operations would be on the means/sums over affs arrays, and that passing in a mask would let scipy skip most of the volume cheaply, but the bincount is a relatively cheap function, and we can make every operation (adding/dividing/comparing) arrays much cheaper by just applying the mask at the start.
Lets go ahead and merge the optimized versions and delete the legacy code

volara/blockwise/extract_frags.py
```diff
@@ -170,22 +197,52 @@ def init_out_array(self):
...
+            # scipy.ndimage.measurements.mean with an index array, which rescans
+            # the whole volume per label. Same values, same order (np.unique is
+            # sorted), one pass.
+            fragment_ids, inverse = np.unique(fragments_data, return_inverse=True)
+            # numpy >= 2.0 returns `inverse` shaped like the input rather than
+            # flat, so ravel explicitly rather than rely on the version default.
+            inverse = inverse.reshape(-1)
+            counts = np.bincount(inverse)
+            sums = np.bincount(inverse, weights=average_affs.reshape(-1))
+            means = sums / counts
```
pattonw: Very nice optimization

volara/blockwise/extract_frags.py
```diff
@@ -170,22 +197,52 @@ def init_out_array(self):
...
+            return fragments_data
 
-        filtered_fragments = np.array(filtered_fragments, dtype=fragments_data.dtype)
         replace = np.zeros_like(filtered_fragments)
-        replace_values(fragments_data, filtered_fragments, replace)
+
+        # `replace_values` builds and returns a new array; it does NOT mutate
+        # `fragments_data`. Discarding this return made `filter_fragments` a
+        # silent no-op.
+        return replace_values(fragments_data, filtered_fragments, replace)
```
pattonw, line 206: this looks like a bug fix. good catch. Surprised we didn't notice this filter failing

volara/blockwise/extract_frags.py
```diff
@@ -231,15 +290,26 @@ def compute_fragments(self, affs_data):
...
+            # Generate the noise straight into `shift` instead of
+            # `zeros_like` + `randn(*shape) * eps` + `+=`, which allocated three
+            # full (C, Z, Y, X) float64 volumes (~2.2 GB each at full context)
+            # to end up with one.
+            shift = np.empty_like(affs_data)
+            if shift.dtype in (np.float32, np.float64):
+                _rng.standard_normal(shift.shape, dtype=shift.dtype, out=shift)
+            else:
+                shift[:] = _rng.standard_normal(shift.shape)
+            shift *= self.noise_eps
```
pattonw, line 320: looks good

volara/blockwise/extract_frags.py
```diff
@@ -349,20 +430,44 @@ def watershed_in_block(
...
+                # A single np.unique plus one bincount per axis replaces both the
+                # separate np.unique(return_counts=True) and
+                # scipy center_of_mass(ones_like(...), ...), which rescanned the
+                # whole volume per label. Identical centroids and sizes.
+                fragment_ids, inverse = np.unique(fragments_data, return_inverse=True)
+                inverse = inverse.reshape(-1)  # numpy >= 2.0 keeps the input shape
+                counts = np.bincount(inverse)
+                logger.info("Found %d fragments", len(fragment_ids))
+
+                centers_of_masses = np.empty(
```
pattonw: bummer that the scipy measurements are so innefficient. So much of the optimizations in here seem to be just rolling a custom version of the scipy measurements which I assumed were already pretty efficient

pattonw (commented):
Overall I'd say we can take pretty much all the optimizations proposed in here. The large majority of them are simply getting around the scipy measurements calls being suboptimal and instead writing custom versions using np.bincount on better masked data. This seems like a pure win. I'd say go through, delete the legacy code, and accept the optimizations

PR #39, opened 2026-07-29
fix(logging): make the log basedir survive the worker process boundary

rhoadesScholar:
## The bug

`set_log_basedir` only mutated a module-level global. Each worker is a fresh `volara-cli`
**process** that re-imports `volara.logging` and therefore gets the default `./volara_logs`,
while the driver uses whatever it set.

That is not merely a misplaced log file. `BlockwiseTask.meta_dir` — and therefore `block_ds`
(`blocks_done.zarr`) — lives under the basedir, so **driver and worker address different
done-marker stores**:

1. the driver's `init_block_array()` creates a proper zarr **Array** at its path;
2. the worker's `mark_block_done_func()` then calls `open_ds(block_ds, mode="a")` on its *own*
   nonexistent path, and zarr v3 with `mode="a"` and no shape creates a **Group**;
3. every subsequent access fails with `Expected a zarr Array ..., got Group`.

The run then **orphans every block while the driver reports nothing wrong** — i.e. silently
incomplete output rather than a crash. Observed on a slabreg composed render as 512/512 blocks
orphaned, with both stores sitting side by side in a single run directory:

```
volara_logs_s1/…/blocks_done.zarr → "node_type": "array"   (driver)
volara_logs/…/blocks_done.zarr    → "node_type": "group"   (worker)
```

## The fix

Back the value with `VOLARA_LOG_BASEDIR`: the env var is read at import so spawned workers
inherit it, and `set_log_basedir` exports it so a driver that changes the basedir at runtime
propagates to workers it spawns afterwards. Default behaviour is unchanged when nobody sets it.

## Verification

- Direct subprocess check: driver sets a basedir → a child `python -c "import volara.logging"`
  reports the same path (previously the default).
- Full suite: **81 passed, 4 skipped, 1 xfailed**.
- End to end: the slabreg composed ROI render that orphaned all 512 blocks now completes.

## Note for reviewers

Any caller combining `set_log_basedir` with multiprocessing is affected, not just slabreg — the
failure signature (all blocks orphaned, no driver-visible cause) is expensive to diagnose, which
is the main argument for landing it.


pattonw:
Why do we need this? Previous jobs have executed fine and worker logs have propagated through to their appropriate locations. And I have used the set_log_dir explicitly before and it has worked as expected.
1) is this a bug due to upgrading to daisy v2? It could be that workers are no longer keeping track of their log directories due to something missed in the daisy upgrade. If so, please make an issue there showing a minimal example of what the behavior was, what it is now, and a simple explanation of how/why thats wrong.
2) If this is a bug that exists in volara itself, please also make a minimum working example demonstrating the error, and then a target pull request that fixes just that

rhoadesScholar:
Closing in favour of **#44**, which is the same fix standalone against `main` — one file, 16 lines, instead of 543/-120 across 12 files. This PR was stacked on #32's branch while targeting `main`, which is why the diff looked like that.

@pattonw — your question (1) was the right one and the answer is yes: it is a **daisy v2 regression**, filed upstream as **funkelab/daisy#75**. daisy 1.x carried `logdir` in the worker `Context` (hence in `DAISY_CONTEXT`, hence across any process boundary); v2 dropped it and substituted a fallback that reads the process-global `get_log_basedir()` — which its own comment describes as what "fork-inheriting workers see as the value the master set." volara spawns `volara-cli` processes rather than forking, so that assumption doesn't hold and the worker silently gets daisy's default.

You were right that it used to work. The mechanism that made it work was removed upstream.

#44 reframes the volara change as a **guard** rather than the fix, and carries the runnable reproduction you asked for plus `tests/test_logging.py` — there was no test crossing a process boundary at all, which is what let this ship. If you'd rather take nothing on the volara side and just wait for daisy#75, that's a reasonable call; the test is worth landing either way.

PR #40, opened 2026-07-30
fix(workers): remove duplicated lines in get_slurm_command / LSF get_command

rhoadesScholar:
## The bug

`SlurmWorker.get_slurm_command` appends the worker command **twice**, so every spawned worker runs:

```
srun ... volara-cli blockwise-worker -c cfg volara-cli blockwise-worker -c cfg
```

and dies with:

```
Error: Got unexpected extra arguments (/path/volara-cli blockwise-worker)
```

daisy reports that per block as `worker function <task> failed`, so **the driver looks healthy while
every block is orphaned**. A full-volume slabreg run lost all 1044 blocks of its first stage twice
before the cause turned up in the worker's own `slurm_worker.err` — the driver-side error is only
`left failed/orphaned blocks`, which points at the data, not the spawn.

## Scope

An unfixed merge artifact (the two nearby "cleanup merge error" commits caught others). In the same
function, also duplicated:

- the `run_command = ["srun"]` initialisation — the **second assignment discards a `--job-name`**
  appended before it
- `--gpus`, `--constraint`, `--output`, `--error`

`srun` tolerates a repeated flag (last wins) so those were survivable, but a doubled `--output`
silently redirects the worker log. `LSFWorker.get_command` had the doubled list init and a doubled
`-J`. Both docstrings had duplicated argument lines.

## After

```
srun --job-name=X --cpus-per-task=4 --gpus=1 --mem=15564 --partition=Q \
     --output=... --error=... <cli> blockwise-worker -c <cfg>
```

CLI, subcommand, `-c`, config path and every flag appear exactly once.

## A note on testing this

Downstream (slabreg) had 6 tests over this path and **all of them passed against the doubled
command** — "contains srun" and "the CLI is pinned" are both true of a duplicated command. A
containment assertion cannot catch duplication. If volara wants a regression test here, it needs to
be exactly-once (`cmd.count("blockwise-worker") == 1`), not membership.

https://claude.ai/code/session_014fHUga9k3ni54vT4mq99vn


pattonw:
Lots of unrelated code on this pull request

pattonw:
again there's too many changes in here for this to be clear what is going wrong and what is being fixed here. Please make a minimal working example showing the error and how this fixes it. And avoid unrelated changes. If you are depending on a different commit (e.g. daisy-v2 support) you can target that branch for merging instead of main so that the diff is clear. If you do that just make it obvious in the description of your pull request

rhoadesScholar:
Closing to make the history honest, rather than landing this as a fix against `main`.

Checking where the duplication actually came from: **it does not exist on `main`.** `main` builds
`sbatch --wrap=<command>` with no duplication at all. The doubling was introduced by the `srun` refactor
on #32's branch (`b412d99`, itself duplicated in history as `fa4e97b` by an earlier merge).

So this was never a bug fix against `main` — it was a fixup to unmerged work, which is exactly why the
diff read as 517/-122 across 12 files for a 28-line deletion, and why @pattonw's *"lots of unrelated code
on this pull request"* was the right call.

Rather than land a "fix" for a bug `main` never had, `c3cf034` will be **squashed into `b412d99`** when the
worker-model refactor is reopened as part of the daisy-v2 chain (see the closing comment on #32 for the
full plan), so the duplication never exists in the history at all.

**The branch is preserved and the item is tracked** — it must not be lost, because merging the daisy-v2
worker refactor without it would reintroduce a spawn path that orphans every block.

## For the record, since the doubled command cost real runs

Constructed both ways with a stub `srun` on PATH — no cluster needed:

```
base    srun --job-name=X --cpus-per-task=4 --gpus=1 --gpus=1 --mem=… --partition=… \
             --output=A --error=B --output=A --error=B \
             volara-cli blockwise-worker -c cfg volara-cli blockwise-worker -c cfg      (19 tokens)

fixed   srun --job-name=X --cpus-per-task=4 --gpus=1 --mem=… --partition=… \
             --output=A --error=B volara-cli blockwise-worker -c cfg                     (12 tokens)
```

## Two corrections to this PR's description, so they don't get carried forward

I overstated two of the duplications, and a reviewer who checked them would rightly have discounted the
rest:

- *"the second assignment discards a `--job-name` appended before it"* — mechanically true, but the second
  block re-appends it, so `--job-name` appears **exactly once** in the output. Net effect: none.
- *"a doubled `--output` silently redirects the worker log"* — the two `--output` values are **identical**,
  so `srun`'s last-wins gives the same path. Net effect: none.

The accurate and stronger framing: **exactly one of the duplications is harmful — the doubled worker
command, which is fatal** (click: *"Got unexpected extra arguments"*, every block orphaned). The rest are
cosmetic.

## The testing point in this PR was right, and still needs acting on

`tests/test_workers.py` **passes against the duplicated command** — verified, 3 passed either way — because
it asserts `cmd[-4:] == [...]` and the duplicate is *appended*, so the tail is still correct. A tail or
containment assertion cannot catch duplication.

The refactor PR will add `cmd.count("blockwise-worker") == 1`. slabreg has already landed exactly that
downstream (`tests/test_slurm_worker.py`), so for now the consumer is better protected than the library.

PR #41, opened 2026-08-03
feat(blockwise): return the daisy TaskState map from run_blockwise on both paths

rhoadesScholar:
Resubmitted from #32, which bundled this feature with the daisy-v2 migration, a worker-model
refactor, a `graph_mws` timeout fix, a `dbs` fix and a ruff pass. @pattonw asked for it to be closed
and reopened with a description that matches the change — this is that. **Nothing here depends on
daisy v2; it targets `main` on daisy 1.x.**

`529/-118 across 11 files` → `233/-8 across 3 files`.

## Problem

`BlockwiseTask.run_blockwise` returns an inconsistent type: `dict[str, TaskState]` on the serial path
but a `bool` on the multiprocessing path (via `daisy.run_blockwise`). Worse, that bool is
`all(task_state.is_done() ...)`, and `TaskState.is_done()` counts failed/orphaned blocks as done — so
a run in which blocks failed still reports `True`, a partial output looks successful, and no caller
can tell.

## Reproduction

Two 10×10 blocks; the lambda raises on the second.

```python
import tempfile, numpy as np
from pathlib import Path
from funlib.geometry import Coordinate
from funlib.persistence.arrays import prepare_ds
from volara.blockwise.lambda_task import LambdaTask
from volara.datasets import Labels, Raw
from volara.logging import set_log_basedir

tmp = Path(tempfile.mkdtemp()); set_log_basedir(tmp / "logs")
# block 0 = all 0.0 -> succeeds; block 1 = all 9.0 -> the lambda raises
data = np.concatenate([np.zeros((10, 10), np.float32), np.full((10, 10), 9.0, np.float32)])
p = tmp / "d.zarr" / "raw"
prepare_ds(p, shape=data.shape, voxel_size=Coordinate(1, 1), dtype=data.dtype, mode="w")[:] = data

task = LambdaTask(
    in_data=Raw(store=p), out_data=Labels(store=tmp / "d.zarr" / "out"),
    lambda_func=lambda x: ((x > 0.5).astype(np.uint8) if x.max() <= 1.5
                           else (_ for _ in ()).throw(RuntimeError("boom"))),
    block_size=Coordinate(10, 10),
)
r = task.run_blockwise(multiprocessing=True)
print(type(r).__name__, "|", r)
```

**On `main`:**
```
bool | True
```
One of two blocks failed. The caller is told `True`.

**On this branch:**
```
defaultdict | {'out-lambda': TaskState(...)}
  total=2 completed=1 failed=1 orphaned=0 is_done()=True
```
Note `is_done()` is still `True` with a failed block — that is exactly why collapsing to
`all(is_done())` hid the failure. `failed_count` is what a caller has to inspect.

## Change

Return daisy's `{task_id: TaskState}` map on **both** paths. The multiprocessing path drives daisy's
`Server` directly via `_run_blockwise_with_states` — a deliberate mirror of
`daisy.convenience._run_blockwise` including the ThreadPool and `stop_event`, so `KeyboardInterrupt`
still stops the server cleanly — instead of the bool-collapsing convenience wrapper. `Pipeline.run_blockwise`
returns the merged map (it previously returned `None`).

Callers can now inspect `failed_count` / `orphaned_count` and react to an incomplete run rather than
silently accepting a partial output. Downstream (slabreg) uses this to raise on any failed or orphaned
block so a partial output never flows into the next stage.

## On fixing it in daisy instead

@pattonw's point on #32 was right, and it still stands: daisy already tracks and returns these states
— `Server.run_blockwise` and `SerialServer.run_blockwise` both return `{task_id: TaskState}`. The only
1.2.2 collapse is `daisy.convenience._run_blockwise` doing `return all(...)`. daisy 2.0 fixes it
properly.

Two things are worth separating, though:
- propagating the map through `BlockwiseTask.run_blockwise` **and** `Pipeline.run_blockwise` is our own
  API surface regardless of daisy — both were discarding what daisy handed back;
- `_run_blockwise_with_states` is the 1.2.2 bridge. It is deliberately isolated so that on daisy 2.0 it
  deletes and becomes a direct call, and the tests pin the return **contract**, not the helper.

## Tests

`tests/test_run_blockwise_states.py`, 4 cases: serial returns the map with the right `completed_count`;
a multiprocessing run whose worker raises returns `failed_count > 0` **while `is_done()` is `True`**
(the exact failure the bool hid); and a `Pipeline` returns the merged map on both paths.

```
$ pytest tests/test_run_blockwise_states.py -q
4 passed
```

Run against `daisy 1.2.2`. Note the import is `from daisy.task_state import TaskState`, which is the
1.x path — daisy 2.0 moves `TaskState` to the top level, so this file needs a one-line change as part
of the v2 migration rather than here.

### DIFF
--- a/tests/test_run_blockwise_states.py
+++ b/tests/test_run_blockwise_states.py
@@ -1,0 +1,176 @@
+"""Tests for the daisy ``{task_id: TaskState}`` map returned by
+``BlockwiseTask.run_blockwise`` and ``Pipeline.run_blockwise``.
+
+Before this behavior existed, the multiprocessing path returned
+``daisy.run_blockwise``'s collapsed bool -- which reports ``True`` even when
+blocks fail, because ``TaskState.is_done()`` counts failed/orphaned blocks as
+"done" -- and ``Pipeline.run_blockwise`` returned ``None``. These tests pin the
+map return on both the serial and multiprocessing paths so a caller can see
+``failed_count`` / ``completed_count`` and react to a partial run.
+"""
+
+import numpy as np
+from daisy.task_state import TaskState
+from funlib.geometry import Coordinate
+from funlib.persistence.arrays import prepare_ds
+
+from volara.blockwise.lambda_task import LambdaTask
+from volara.datasets import Labels, Raw
+
+
+def _write_raw(path, data):
+    prepare_ds(
+        path,
+        shape=data.shape,
+        voxel_size=Coordinate(1, 1),
+        dtype=data.dtype,
+        mode="w",
+    )[:] = data
+    return path
+
+
+# A lambda that raises when a block's data exceeds a threshold. Passed through
+# volara's cloudpickle-based PydanticCallable, so it must be an expression --
+# hence the generator ``.throw`` idiom to raise inside the conditional.
+_FAIL_ON_HIGH = lambda x: (  # noqa: E731
+    (x > 0.5).astype(np.uint8)
+    if x.max() <= 1.5
+    else (_ for _ in ()).throw(RuntimeError("boom: block deliberately failed"))
+)
+
+
+def test_task_serial_returns_state_map(tmp_path):
+    """multiprocessing=False returns {task_id: TaskState}; all blocks complete.
+
+    pytest tests/test_run_blockwise_states.py::test_task_serial_returns_state_map
+    """
+    # 20x10 tiled into two 10x10 blocks
+    data = np.linspace(0, 1, 200, dtype=np.float32).reshape(20, 10)
+    in_path = _write_raw(tmp_path / "data.zarr" / "raw", data)
+
+    task = LambdaTask(
+        in_data=Raw(store=in_path),
+        out_data=Labels(store=tmp_path / "data.zarr" / "out"),
+        lambda_func=lambda x: (x > 0.5).astype(np.uint8),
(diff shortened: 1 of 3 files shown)


rhoadesScholar:
Closing as superseded by the daisy-v2 migration line.

This PR's three commits are exact cherry-picks (equal `git patch-id --stable`) of the bottom three commits of `feat/run-blockwise-states` (`0ce19c6`, `c895720`, `607f052`), now carried by `will/rbws-plus-main` merged with `main`. That branch implements the same `{task_id: TaskState}` return contract on both paths — identical docstrings, and a `tests/test_run_blockwise_states.py` whose body is byte-identical to this PR's — via daisy 2.0's native `return_states`, which supersedes the daisy-1.x `_run_blockwise_with_states` helper this PR needed.

Verified before closing: merging this branch into `will/rbws-plus-main` conflicts in 3 files, and an ours-side resolution produces a tree byte-identical to that branch's tip (`bd89ad7`) — this PR contributes no content the migration line does not already have.

One consequence, deliberate: `main` keeps the old contract (`bool` from the multiprocessing path, `None` from `Pipeline.run_blockwise`) until the migration merges. Downstream consumers that need the map (e.g. slabreg's completion guard) are repinning to the migration line instead.

The head branch is deleted; these commits remain reachable via `refs/pull/41/head`.

Issue #34, opened 2026-07-22
Migrate to daisy v2 (Rust rewrite) once released — API exposure audit


rhoadesScholar:
## Context

daisy v2 ([`funkelab/daisy` branch `v2.0`](https://github.com/funkelab/daisy/tree/v2.0)) is a complete rewrite with a Rust scheduling core and PyO3 bindings. It is **not on PyPI yet** (latest published: 1.2.2), but its own changelog says *"If you depend on the daisy 1.x API surface, pin `daisy<2`"* — and volara's current `daisy>=1.2.2` (uncapped) will auto-upgrade and break CI (`--resolution highest`) the moment v2.0.0 ships. A `daisy>=1.2.2,<2` pin is being added in #32; this issue tracks the actual migration.

This audit is against the `v2.0` branch as of 2026-07-22 (`MIGRATION.md`, `CHANGELOG.md`, `daisy-py/python/daisy/*`, `daisy-py/python/daisy/_daisy.pyi`).

## Hard breaks (imports that fail on v2)

| volara usage | Where | v2 replacement |
|---|---|---|
| `from daisy.cl_monitor import CLMonitor` | `blockwise/blockwise.py`, `blockwise/pipeline.py` | Module removed. Use `run_blockwise(progress=...)` / `JsonProgressObserver`; the tqdm observer is the default. |
| `from daisy.tcp import IOLooper` | `blockwise/blockwise.py` (`_run_blockwise_with_states`, from #32) | Module removed. **The whole helper gets deleted**: v2 `Server().run_blockwise(pipeline)` natively returns the `{task: TaskState}` map, and serial runs via `run_blockwise(..., multiprocessing=False)` also return states. |
| `from daisy.block import BlockStatus` | `blockwise/blockwise.py` | Submodule path gone; import top-level `daisy.BlockStatus`. |
| `daisy.get_worker_log_basename(...)` | `workers.py` (3×) | No longer re-exported at top level; use `daisy.logging.get_worker_log_basename`. |
[...]

PR #42, opened 2026-08-03
deps(daisy): cap daisy<2 until the v2 migration lands

rhoadesScholar:
Split out of #32 as a standalone, one-line change. #32 went straight to a git pin on daisy's
unreleased `v2.0` branch, so the defensive cap that was step 1 of #34 never landed and `main` is
still exposed.

## Why

`main` declares `daisy>=1.2.2` with no upper bound, so a daisy 2.0.0 release on PyPI resolves for
every fresh install of volara `main` — and for every downstream repo that says `daisy>=1.2.2`
(volara-surface, volara-registration, frag_embed, slicing-metric all do).

**It does not fail loudly, which is the actual problem.** volara `main` imports fine against daisy
2.0.0 and the whole suite passes:

```
$ pip install 'daisy==2.0.0' && pytest tests -q
163 passed, 3 skipped, 2 xfailed
```

daisy v2's `v1_compat` layer still provides `cl_monitor`, `get_worker_log_basename`, `SerialServer`
and friends, so nothing in #34's import-break list actually raises. What changes is behaviour,
silently:

```python
>>> import daisy
>>> daisy.Context(hostname="h", port=1, task_id="t", worker_id=0).to_env()
```
| daisy | result |
|---|---|
| 1.x | `hostname=h:port=1:task_id=t:worker_id=0:logdir=<the driver's basedir>` |
| **2.0** | `hostname=h:task_id=t:port=1:worker_id=0` — **no `logdir`** |

`volara/blockwise/blockwise.py:260` reads `client.context["logdir"]` to recover the driver's log
basedir inside the worker. Under v2 that key is absent from the wire context, and v2's `Client.__init__`
re-injects it from the process-global `get_log_basedir()` — which its own comment describes as the value
*"fork-inheriting workers see as the value the master set."*

**volara's workers are spawned `volara-cli` processes, not forks.** So the global is daisy's default,
not the driver's. There is no `KeyError` and no crash — the worker silently resolves a *different*
basedir from the driver.

Since `BlockwiseTask.meta_dir` (and therefore `block_ds` / `blocks_done.zarr`) lives under the basedir,
driver and worker then address **different done-marker stores**: the driver's `init_block_array()`
creates a zarr Array, the worker's `open_ds(block_ds, mode="a")` creates a **Group** at its own
nonexistent path, and every block is orphaned while the driver log stays clean.

That failure has already cost real full-volume runs downstream. A cap is one line and it decouples the
daisy-2.0.0-release deadline from the migration work.

## Scope

```diff
-    "daisy>=1.2.2",
+    "daisy>=1.2.2,<2",  # daisy 2.0 is a breaking Rust rewrite; see e11bio/volara#34
```

No behaviour change on daisy 1.x, so there is no failing→passing example to show — the evidence above
is the *reason*, and the change itself is a constraint, not code. Remove the cap as part of the v2
migration.

Refs #34.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ requires-python = ">=3.11"
 
 dependencies = [
     "click>=8.0",
-    "daisy>=1.2.2",
+    "daisy>=1.2.2,<2",  # daisy 2.0 is a breaking Rust rewrite; see e11bio/volara#34
     "funlib.geometry>=0.3",
     "funlib.persistence>=0.7.0",
     "mwatershed>=0.5.2",


pattonw:
This looks like a bandaid for a bug on daisy v2. daisy v2 should be a perfect drop in replacement for v1. So if there is a difference in the environment (logdir missing), that should be a pull request on daisy v2.

For now I'll go ahead and merge this since until we are happy enough with daisy v2 we want to avoid installing in on main unless explicitly desired

PR #56, opened 2026-08-04
feat(benchmark): trace normal runs, and drop sqlite and polars from the benchmark path

rhoadesScholar:
## What this is

An overhaul of volara's benchmarking, replacing the patch this PR used to be.
@pattonw's [review](https://github.com/e11bio/volara/pull/56#issuecomment-5185545231)
called the previous version "patching over some poor design decisions in the
original benchmarking implementation" and raised three points; this answers all
three, and fixes the bug the patch was originally for.

## The bug underneath

`BlockwiseTask.get_benchmark_logger` is the **only** way any code in volara gets
a `BenchmarkLogger` — every `trace()` in `blockwise.py`, `extract_frags.py`,
`aff_agglom.py`, `graph_mws.py` and `relabel.py` goes through it. On `main` it
reads:

```python
def get_benchmark_logger(self) -> BenchmarkLogger:
    _benchmark_db_path = Path("volara_benchmark_logs/benchmark.db")   # computed...
    return BenchmarkLogger(
        None,                                                        # ...and discarded
        task=self.task_name,
    )
```

With `db_path=None` the logger's connection is `None` and both `log()` and
`trace()` early-return, so no writer in the codebase ever wrote a row.
`benchmark()` then called `print_report()` on a table nobody had written to,
which raised out of a `finally` block.

## What changed

| | `main` | this branch |
|---|---|---|
| what `benchmark()` runs | always a spoofed dry run | a **real run**, traced. `spoof=True` for the dry run |
| trace sink | one sqlite db, every worker writing to it | one append-only `.jsonl` per process, reduced in the driver |
| report | polars | `statistics` + `csv` — `polars` removed from `pyproject.toml` |
| per-trace cost | a fresh `daisy.Client()`, i.e. a TCP connect to the scheduler | an `os.environ` read |
| `benchmark()` + `worker_config` | broken under daisy 2.0 (below) | works; pinned by a test **under daisy 1.2.2 only** — `pyproject.toml` pins `daisy>=1.2.2,<2`, so CI cannot reach the 2.0.0 path where the log-basedir regression lives; that half was verified by hand, not by CI |
| empty report | `InvalidOperationError` out of a `finally` | prints an empty report |

**(a) the common case.** Benchmarking a job as it runs normally, the first time,
is what `benchmark()` now does with no arguments: real outputs, real block done
dataset, already-done blocks still skipped, ordinary log basedir. Spoofing
outputs so that an already-run or mid-run task can be benchmarked without
disturbing its data is still there as `benchmark(spoof=True)` — including
dropping every artifact afterwards. Same for `Pipeline.benchmark`.

**(b) sqlite.** Gone. Each process appends to a trace file of its own under
`volara_benchmark_logs/traces/`, one JSON record per traced operation; the
driver globs and reduces them at the end. Nothing is shared between writers, so
there is no lock to contend on — the same "one writer per key" property as the
block done zarr. Files rather than a zarr because a trace is not one-per-block:
there is `init` per task, and nested per-block operations (`Read Affs`,
`Compute Fragments`, `Write Fragments`, …), so there is no block-indexed array
to write into without inventing an operation axis.

**(c) polars.** Gone from the benchmark path *and* from `pyproject.toml` —
`print_report` was its only user anywhere in volara. Aggregation is
`statistics.fmean`/`stdev`, output is `csv.writer`, and the formatting is
unchanged. Worth noting the zombie-process symptom is at least partly retired
already: `test_benchmark_traces_the_worker_loop` runs `multiprocessing=True` and
passed on all six test legs before this change too.

### `benchmark()` + `worker_config`, and a correction

My earlier note on this PR said this path fails. Correcting that: on **daisy
1.2.2** — what `pyproject.toml` resolves and what CI runs — I could not
reproduce it; 2/2 blocks complete and rows land. It reproduces on **daisy
2.0.0**, where `Client.__init__` no longer puts `logdir` in the worker context
and falls back to the *worker's* own log basedir. A freshly spawned `volara-cli
blockwise-worker` therefore resolved `volara_logs/<task>-meta/blocks_done.zarr`
while the scheduler had created it under the benchmark basedir:

```
TypeError: Expected a zarr Array at volara_logs/spoof_out-lambda-meta/blocks_done.zarr, got Group
```

— every block failed while `benchmark()` still returned and still wrote a
report. A benchmark run now exports its log basedir and workers adopt it
directly, which is right under both versions.
`test_benchmark_traces_a_worker_config_run` pins it, and asserts on the
`Process Block` count rather than on the run completing, precisely because
completing is not evidence.

## Reproduction

Same script both sides. It uses `worker_config=LocalWorker()`, i.e. the
distributed path, since that is the one that has to work.

```python
# bench_repro.py -- usage: python bench_repro.py <workdir>
import os
import sys
from pathlib import Path

import numpy as np
from funlib.geometry import Coordinate
from funlib.persistence.arrays import prepare_ds

from volara.blockwise import LambdaTask
from volara.datasets import Labels, Raw
from volara.workers import LocalWorker

work = Path(sys.argv[1]).absolute()
work.mkdir(parents=True, exist_ok=True)
os.chdir(work)
data = np.linspace(0, 1, 200, dtype=np.float32).reshape(20, 10)
raw = work / "data.zarr" / "raw"
prepare_ds(raw, shape=data.shape, voxel_size=Coordinate(1, 1),
           dtype=data.dtype, mode="w")[:] = data

task = LambdaTask(
    in_data=Raw(store=raw, writable=False),
    out_data=Labels(store=work / "data.zarr" / "out"),
    lambda_func=lambda x: (x > 0.5).astype(np.uint8),
    block_size=Coordinate(10, 10),   # -> 2 blocks
    num_workers=2,
    worker_config=LocalWorker(),     # the distributed path
)

try:
    task.benchmark(multiprocessing=True)
    print("benchmark() returned cleanly")
except Exception as e:
    print(f"benchmark() raised {type(e).__name__}: {str(e).splitlines()[0]}")

print("output written by the run:", (work / "data.zarr" / "out").exists())
for csv in sorted(Path("volara_benchmark_report").glob("*.csv")):
    print(f"--- {csv.name} ---")
    print(csv.read_text(), end="")
```

**On `main` (66106bb), daisy 1.2.2:**

```
Symlinking .../data.zarr/raw
Spoofing .../data.zarr/out

Execution Summary
-----------------

  Task spoof_out-lambda:

    num blocks : 2
    completed ✔: 2 (skipped 0)
    failed    ✗: 0
    orphaned  ∅: 0

    all blocks processed successfully
benchmark() raised InvalidOperationError: UDF called without return type, but was not able to infer the output type.
output written by the run: False
```

Both blocks ran and **nothing was recorded** — the report does not even come out
empty, it raises, because polars cannot infer types over a frame built from zero
rows.

**On this branch, daisy 1.2.2:**

```
Execution Summary
-----------------

  Task out-lambda:

    num blocks : 2
    completed ✔: 2 (skipped 0)
    failed    ✗: 0
    orphaned  ∅: 0

    all blocks processed successfully
task        init                         Process Block Setup            Acquire Block                  Process Block                  Mark Block Done
out-lambda  0.032s ± 0.0 (idle: 0.002s)  0.008s ± 0.001 (idle: 0.003s)  0.009s ± 0.005 (idle: 0.009s)  0.006s ± 0.001 (idle: 0.001s)  0.004s ± 0.0 (idle: -0.001s)
Benchmark report written to volara_benchmark_report
benchmark() returned cleanly
output written by the run: True
--- io.csv ---
task,init,Process Block Setup,Acquire Block,Process Block,Mark Block Done
out-lambda,read/write: 0.0/0.02 MB,read/write: 0.0/0.0 MB,read/write: 0.0/0.0 MB,read/write: 0.0/0.0 MB,read/write: 0.0/0.0 MB
--- memory.csv ---
task,init,Process Block Setup,Acquire Block,Process Block,Mark Block Done
out-lambda,0.63 MB,3.02 MB,0.07 MB,0.45 MB,0.36 MB
--- time.csv ---
task,init,Process Block Setup,Acquire Block,Process Block,Mark Block Done
out-lambda,0.032s ± 0.0 (idle: 0.002s),0.008s ± 0.001 (idle: 0.003s),0.009s ± 0.005 (idle: 0.009s),0.006s ± 0.001 (idle: 0.001s),0.004s ± 0.0 (idle: -0.001s)
```

Note `output written by the run: True` — this was a real run of the task, not a
dry run, and the worker-loop operations (`Acquire Block`) are in the report, so
the separate `volara-cli` interpreters were traced too.

## Tests

`tests/test_benchmark.py`, 12 tests. Run with `uv`-resolved deps, i.e. daisy
1.2.2, matching CI:

```
$ pytest tests/test_benchmark.py -q -p no:warnings
............                                                             [100%]
12 passed in 9.99s
```

Full suite, no regressions:

```
origin/main (66106bb):  177 passed, 4 skipped, 2 xfailed in 27.05s
this branch:            189 passed, 4 skipped, 2 xfailed in 36.84s
```

The regression that matters,
`test_benchmark_traces_a_worker_config_run[True]`, fails without the log-basedir
fix under daisy 2.0.0 (verified by reverting just that hunk):

```
$ pytest tests/test_benchmark.py::test_benchmark_traces_a_worker_config_run -q -p no:warnings
TypeError: Expected a zarr Array at volara_logs/spoof_out-lambda-meta/blocks_done.zarr, got Group
tests/test_benchmark.py:186: AssertionError: blocks did not run in the worker; recorded [...]
FAILED tests/test_benchmark.py::test_benchmark_traces_a_worker_config_run[True]
1 failed, 1 passed in 11.24s
```

`ruff check volara tests` passes and the four touched files are
`ruff format`-clean.

## Not in this PR

- `Dataset.spoof` rejects `str` stores though `store` is typed `Path | str`, and
  `SQLite.spoof(nodes: bool = True)` is called as `value.spoof(spoof_dir)` — the
  directory arrives as `nodes`. Both block `spoof=True` on the CREMI example;
  see #55. The default (non-spoof) path is unaffected.
- `LUT` has no `spoof`, so `GraphMWS` writes the real lut during a `spoof=True`
  run.
- `examples/benchmarks/benchmark_cremi.py` is untouched; its
  `pipeline.benchmark(multiprocessing=True)` now runs for real rather than
  spoofing, which is the only way it can work at all until #55 lands.
- Three `tests/test_logging.py` tests fail under daisy 2.0.0; they fail on
  `main` under 2.0.0 too and are untouched here.

## Skeptical pass

One bucket: the benchmarking subsystem. The `polars` removal from
`pyproject.toml` is in-bucket because `print_report` was its only consumer — the
dependency exists solely for the code this PR replaces. Nothing else in the diff
is unrelated: `blockwise.py` changes are `get_benchmark_logger`, `benchmark()`,
and the worker log basedir; `pipeline.py` mirrors `benchmark()`. `run_blockwise`
behaviour is unchanged, and `test_run_blockwise_stays_unbenchmarked` and
`test_benchmark_does_not_leak_the_signal` exist to keep it that way.

Left as a draft deliberately.

---

**Update 2026-08-05 — one more fix, and one correction to the table above.**

`benchmark()` called twice in the same interpreter and cwd raised `FileNotFoundError`:
`benchmark_run` wipes `volara_benchmark_logs/traces` on every run, while `_process_trace_file`
caches its handle on `(pid, trace_dir)` — on the second call both matched, so the cache hit skipped
the `mkdir` and returned a path whose directory had just been removed. No test caught it because
every test gets a fresh `tmp_path` cwd, so the cache always missed. Fixed in `a2f14e2`
(reset-on-wipe plus an unconditional `mkdir`), with
`tests/test_benchmark.py::test_benchmark_twice_in_one_process` pinning it — `FileNotFoundError`
before, passing after. Full suite **190 passed, 4 skipped, 2 xfailed**; `ruff` clean.

The correction: the `worker_config` row said "pinned by a test" without qualification. The
`db_path=None` half genuinely is pinned (reverting it fails 6 of 13 tests). The daisy-2.0.0
log-basedir half is **not reachable by CI** — `pyproject.toml` pins `daisy>=1.2.2,<2` — and was
verified by hand. Disabling that hunk leaves every test green, so please read it as a hand-verified
fix, not a CI-guarded one.

### DIFF
--- a/volara/blockwise/blockwise.py
+++ b/volara/blockwise/blockwise.py
@@ -19,7 +19,12 @@ from volara.logging import get_log_basedir, set_log_basedir
 
 from ..utils import PydanticRoi, StrictBaseModel
 from ..workers import Worker
-from .benchmark import BenchmarkLogger
+from .benchmark import (
+    BenchmarkLogger,
+    benchmark_run,
+    get_benchmark_dir,
+    get_benchmark_log_basedir,
+)
 
 if TYPE_CHECKING:
     from .pipeline import Pipeline
@@ -257,6 +262,14 @@ class BlockwiseTask(StrictBaseModel, ABC):
         Start our workers and run through every block until a task
         is complete.
         """
+        benchmark_log_basedir = get_benchmark_log_basedir()
+        if benchmark_log_basedir is not None:
+            # A benchmark run may relocate the log basedir, and with it this
+            # task's meta dir and block-done array. Adopt it before anything
+            # resolves a path off it -- daisy's worker context cannot be relied
+            # on to carry it (see BENCHMARK_LOG_BASEDIR_ENV_VAR).
+            set_log_basedir(benchmark_log_basedir)
+
         benchmark_logger = self.get_benchmark_logger()
         with ExitStack() as stack:
             with benchmark_logger.trace("Process Block Setup"):
@@ -265,11 +278,13 @@ class BlockwiseTask(StrictBaseModel, ABC):
             def worker_loop():
                 worker_benchmark_logger = self.get_benchmark_logger()
                 client = daisy.Client()
-                # TODO: this shouldn't be necessary, daisy should be doing this for us
-                try:
-                    set_log_basedir(client.context["logdir"])  # type: ignore[non-subscriptable]
-                except KeyError as e:
-                    raise ValueError(client.context) from e
+                if benchmark_log_basedir is None:
+                    # TODO: this shouldn't be necessary, daisy should be doing
+                    # this for us
+                    try:
+                        set_log_basedir(client.context["logdir"])  # type: ignore[non-subscriptable]
+                    except KeyError as e:
+                        raise ValueError(client.context) from e
                 mark_block_done = self.mark_block_done_func()
 
                 while True:
@@ -408,9 +423,17 @@ class BlockwiseTask(StrictBaseModel, ABC):
             yield task
 
     def get_benchmark_logger(self) -> BenchmarkLogger:
-        _benchmark_db_path = Path("volara_benchmark_logs/benchmark.db")
+        """
+        The logger every ``trace``/``log`` call in this codebase goes through.
+
+        Returns a live logger only while a :meth:`benchmark` run is in progress
+        (signalled by ``VOLARA_BENCHMARK_DIR``, which ``benchmarking()`` exports
+        so that worker processes see it too). Outside of a benchmark run
+        ``get_benchmark_dir()`` is ``None`` and every ``trace``/``log`` becomes a
+        no-op -- ordinary ``run_blockwise`` executions stay untraced.
+        """
         return BenchmarkLogger(
-            None,
+            get_benchmark_dir(),
             task=self.task_name,
         )
 
@@ -428,48 +451,48 @@ class BlockwiseTask(StrictBaseModel, ABC):
                 data[name] = value
         return self.__class__(**data)
 
-    def benchmark(self, multiprocessing: bool = True) -> dict:
+    def benchmark(
+        self,
+        multiprocessing: bool = True,
+        spoof: bool = False,
+        out_dir: Path | None = None,
+    ) -> dict:
         """
-        A helper function for benchmarking and debugging a blockwise task or pipeline.
+        Run this task with tracing switched on and write a timing, memory and
+        io report.
 
-        Used as a "dry run" of `run_blockwise` without saving any outputs.
+        By default this is an ordinary :meth:`run_blockwise` with tracing added:
+        real outputs, real block done dataset, already-completed blocks still
+        skipped. That is the common case -- benchmarking a job as it runs
+        normally, the first time.
 
-        - Will not skip blocks that have already been processed. You can benchmark a task
-            that has already been run.
-        - Will not save any run artifacts such as block done datasets, output datasets,
-            graph nodes or edges, or look up tables. The only thing that will be saved are the
-            worker logs and a timing report.
+        Args:
+            spoof: run against spoofed outputs instead, dropping every artifact
+                afterwards, so that a task which has already run (or is running)
+                can be benchmarked without disturbing the data it produced. No
+                block is skipped in this mode: the spoofed task has its own,
+                empty, block done dataset.
+            out_dir: where to write the report. Defaults to
+                ``./volara_benchmark_report``.
         """
-        from volara.logging import set_log_basedir
+        with benchmark_run(out_dir=out_dir, relocate_logs=spoof):
+            run_self = (
+                self.spoof(Path("volara_benchmark_logs/spoof")) if spoof else self
+            )
 
-        log_basedir = get_log_basedir()
-        set_log_basedir("volara_benchmark_logs")
-        benchmark_db_path = Path("volara_benchmark_logs/benchmark.db")
-        if benchmark_db_path.exists():
-            benchmark_db_path.unlink()
-        benchmark_logger = BenchmarkLogger(task=None, db_path=benchmark_db_path)
-        benchmark_logger._init_db()
+            try:
+                with run_self.task(multiprocessing=multiprocessing) as task:
+                    tasks = [task]
+                    if multiprocessing:
+                        result = daisy.run_blockwise(tasks)  # noqa
+                    else:
+                        server = daisy.SerialServer()
+                        _cl_monitor = CLMonitor(server)
+                        result = server.run_blockwise(tasks)
 
-        spoof_dir = Path("volara_benchmark_logs/spoof")
-        debug_self = self.spoof(spoof_dir)
-
-        try:
-            with debug_self.task(multiprocessing=multiprocessing) as task:
-                tasks = [task]
-                if multiprocessing:
-                    result = daisy.run_blockwise(tasks)  # noqa
-                else:
-                    server = daisy.SerialServer()
-                    _cl_monitor = CLMonitor(server)
-                    result = server.run_blockwise(tasks)
-
-        except Exception as e:
-            raise e
-
-        finally:
-            debug_self.drop()
-            benchmark_logger.print_report()
-            set_log_basedir(log_basedir)
+            finally:
+                if spoof:
+                    run_self.drop()
 
         return result
 
--- a/volara/blockwise/blockwise.py
+++ b/volara/blockwise/blockwise.py
@@ -268,6 +268,11 @@ class BlockwiseTask(StrictBaseModel, ABC):
             # task's meta dir and block-done array. Adopt it before anything
             # resolves a path off it -- daisy's worker context cannot be relied
             # on to carry it (see BENCHMARK_LOG_BASEDIR_ENV_VAR).
+            # This is early enough because it runs at the entry point of the only
+            # process that needs it: a freshly spawned worker (volara-cli
+            # blockwise-worker -> process_blocks), which never runs task()/init.
+            # init traces driver-side, where benchmark_run() has already set the
+            # basedir (and exported both env vars) before task() is entered.
             set_log_basedir(benchmark_log_basedir)
 
         benchmark_logger = self.get_benchmark_logger()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,6 @@ dependencies = [
     "scipy>=1.15.3",
     "scikit-image>=0.25.2",
     "cloud-volume>=12.4.1",
-    "polars>=1.31.0",
     "cloudpickle>=3.0",
     "psutil>=7.0.0",
     # not a direct dependency, but we have to specify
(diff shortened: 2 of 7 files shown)


pattonw:
Thanks for working on this Jeff. I think this is patching over some poor design decisions in the original benchmarking implementation.
1) spoofing data: I specifically wanted to be able to benchmark tasks that had already run or were mid run without messing with the data that was already produced. This lead to spoofing the outputs of each task. Since we already specify outputs for dropping output artifacts this was doable. But this use case has been rare and the more common use case of just wanting to benchmark a job as it runs normally for the first time is much more common and not supported under the current implementation.
2) writing to sqlite db. I thought an sqlite db would be sufficient for tracking timing logs, but I found that it very quickly got locked and caused problems. We need a better solution, and I'm starting to lean more towards something like block done zarrs.
3) pandas/polars dataframe management. pandas is a big heavy dependency that I didn't particularly want to pull in. polars seemed better but led to excessive zombie processes that refused to shut down. It seems polars (at least the version I was using) did not play well with multiprocessing.

So if you want to use the benchmarking, you are free to overhaul it, especially relating to any of the above issues. I'm open to ideas for doing benchmarking in a more stable way

rhoadesScholar:
Thanks @pattonw — that is the answer I was hoping for. Doing the overhaul; I've put #56 back to draft so nobody reviews the design that's being replaced.

**Design**

- **A normal run is the primary path.** `benchmark()` now runs the task for real — same outputs, same block-done cache, same resume semantics as `run_blockwise` — and only adds tracing. Your original case survives as opt-in `benchmark(spoof=True)`: spoofed outputs, artifacts dropped afterwards, for benchmarking something already run or mid-run. Same for `Pipeline.benchmark`.
- **Sink: one append-only JSONL file per process under `volara_benchmark_logs/traces/`, reduced in the driver.** No sqlite, no lock. Each process owns exactly one file and each record is a single `O_APPEND` line write, so N workers never contend — the same "one writer per key" property the block-done zarr gets you. I went with files rather than a zarr because a trace isn't one-per-block: there's `init` per task, and nested per-block operations (`Read Affs`, `Compute Fragments`, `Write Fragments`, …), so there's no block-indexed array to write into without inventing an operation axis.
- **Report: stdlib.** `statistics` + `csv`, same three CSVs and same formatting as now. `polars` comes out of `pyproject.toml` entirely — `print_report` was its only user anywhere in volara. An empty run prints an empty report instead of raising, and `print_report` now actually prints the timing table rather than only writing files.
- **`trace()` no longer constructs a `daisy.Client()`.** It did, on *every* trace, which opens a TCP connection to the scheduler. Worker id now comes from `daisy.Context.from_env()` — parse only, no connection.
- **`benchmark()` + `worker_config` is fixed.** Correcting my own earlier note here: on `daisy` 1.2.2 — what `pyproject.toml` resolves and what CI runs — I could not reproduce a failure; 2/2 blocks complete and rows land. It reproduces on `daisy` 2.0.0, where `Client.__init__` no longer carries `logdir` in the worker context and falls back to the *worker's* own log basedir, so a freshly spawned `volara-cli blockwise-worker` resolves `volara_logs/<task>-meta/blocks_done.zarr` and every block dies with `Expected a zarr Array …, got Group`. A benchmark run now hands its log basedir to its workers explicitly instead of reading it back out of daisy's context, which is correct under both versions. There's a test pinning `benchmark()` + `worker_config` actually recording `Process Block` rows.

On (3) specifically: it's at least partly retired already — `test_benchmark_traces_the_worker_loop` on this branch runs `multiprocessing=True` and passes on all six test legs. Dropping polars removes what's left of the exposure.

I'll push the overhaul here and leave it draft.

volara/blockwise/blockwise.py
```diff
@@ -257,6 +262,14 @@ def process_blocks(self):
         Start our workers and run through every block until a task
         is complete.
         """
+        benchmark_log_basedir = get_benchmark_log_basedir()
```
pattonw, line 265: Is this early enough? Doesn't init also need to get the benchmark log basedir?

pattonw (commented):
Looks pretty good. I just left one comment

volara/blockwise/blockwise.py
rhoadesScholar: Good question — it's early enough, because `init` never reaches this code. The split is by process:

- **`init` traces driver-side only**: `benchmark()` / `Pipeline.benchmark()` enter `benchmark_run()`, which calls `set_log_basedir(...)` and exports both env vars **before** `task()` is entered — so `trace("init")` and `init_block_array()` both resolve against the benchmark basedir already.
- **This adoption exists for freshly spawned workers** (`volara-cli blockwise-worker` → `process_blocks`, cli.py:35): a new process whose module global is unset and whose daisy-2.0 context carries no `logdir`. That path never runs `task()`/`init`.

Verified the third path too: `Pipeline.benchmark` wraps in `benchmark_run()` before any `node.task(...)`. I've added a comment at this line stating the invariant (60e7557) so it doesn't need re-deriving; `tests/test_benchmark.py` is 13/13 on the branch.

PR #58, opened 2026-08-05
feat(daisy-v2): migrate to daisy 2.0 — feat/run-blockwise-states converged with main

rhoadesScholar:
Opens a PR for @pattonw's convergence branch so the daisy-v2 migration line has an owner and a merge trigger. The branch is `feat/run-blockwise-states` (`bccbf78`) merged with `main` (`66106bb`): it contains **100% of both sides** and is 0 behind `main`.

## What this carries

- The daisy 2.0 migration: `daisy @ git v2.0` (Rust rewrite), native `{task_id: TaskState}` returns from `run_blockwise` on both paths via `return_states`, native Roi/Block boundary, `block_timeout` field.
- The worker refactor: blocking `srun` / `bsub -K` submissions, plus #57's fix for the doubled worker command.
- Everything on `main` through #54.

## Supersedes

- #32 (closed 2026-08-03; this branch is its continuation),
- #41 (closed 2026-08-05 — its three commits are patch-identical to this branch's bottom three; merging it here was verified to be a byte-identical no-op),
- `feat/run-blockwise-states` itself (deleted 2026-08-05; fully contained here).

## Verified locally, 2026-08-05

- Containment: `git rev-list --count <side> ^will/rbws-plus-main` → 0 for both `bccbf78` and `66106bb`.
- Conflict resolutions audited file-by-file: `pyproject.toml`, `blockwise.py`, `pipeline.py`, `graph_mws.py` are correct unions — #53, #54, #46, #47 all present; unmerged #45 correctly absent.
- Test suite on this tip (py3.13 venv, daisy 2.0.0 built from `v2.0`): **180 passed, 4 failed, 4 skipped (no Postgres), 2 xfailed**. Of the 4: `test_lambda_task_multiprocessing_local_worker` fails identically on `bccbf78` (pre-existing); the 3 `test_logging.py` failures are likely an artifact of the local daisy build (`26f10edb`, 2026-07-24) predating funkelab/daisy@`505e7be` (logdir carried in worker context) — needs a re-run against a fresh `v2.0` resolve before being treated as real.

## Known issues on this branch

1. `extract_frags.py:17` — unused `measurements` import resurrected by the merge; breaks the `ruff` job (both parents pass). Suggestion posted inline.
2. The daisy pin floats on `branch = "v2.0"` (`pyproject.toml:25` and `:84`). Recommend SHA-pinning `b323d78` — the current tip, which carries `7128d9c` (the loopback fix that broke every remote worker).
3. `.github/workflows/ty.yaml:15` pins uv 0.4.1, which cannot parse this branch's dual daisy declaration — `ty` red is an install failure here, not diagnostics. One-line workflow fix.

## Merge gate

daisy 2.0.0 on PyPI (PyPI rejects `daisy @ git+…` in a published release). Until then this PR is the reviewable, pinnable home of the migration — slabreg is repinning to this branch's tip.


volara/blockwise/extract_frags.py
```diff
@@ -9,8 +9,13 @@
...
 from pydantic import Field
-from scipy.ndimage import gaussian_filter, label, maximum_filter
-from scipy.ndimage.morphology import distance_transform_edt
+from scipy.ndimage import (
+    distance_transform_edt,
+    gaussian_filter,
+    label,
+    maximum_filter,
+    measurements,
+)
```
rhoadesScholar, line 18: The merge resurrected `measurements`, which `main` deleted when both its uses were replaced with `np.bincount` (this is the only diff between this file and `main`'s — everything else is byte-identical). It has zero uses on this branch (`git grep -n measurements` → only this line) and fails the `ruff` job with `F401`; both parents pass.

```suggestion
from scipy.ndimage import (
    distance_transform_edt,
    gaussian_filter,
    label,
    maximum_filter,
)
```

rhoadesScholar:
**Correcting my own claim in the PR body — the three `test_logging.py` failures are real, not a stale-build artifact.**

I hedged that they might come from the local daisy build (`26f10eb`, 2026-07-24) predating funkelab/daisy@`505e7be`. CI disproves that: it resolves the current `v2.0` tip (`b323d78`, which *does* contain `505e7be`) and fails identically — `KeyError: 'logdir'` at `tests/test_logging.py:55`. All six test legs are red for this reason.

**The actual diagnosis.** daisy 2.0 still delivers `logdir` to workers, but only through its *spawn* path: `context_with_logdir()` injects it when `DAISY_CONTEXT` is built for a spawned worker (`daisy-py/python/daisy/_worker_processes.py:118-144`, used at `:247`), and `Client` falls back to its own `get_log_basedir()` otherwise (`_task.py:338-341`). daisy 1.x instead put it on *every* `Context` at construction. volara's test asserts the 1.x form — that a `Context()` built in the driver process already carries `logdir` — so it fails under v2 by design, not by regression.

The contract the test protects is still honoured: daisy's own `tests/test_worker_execution.py:249` asserts a spawned worker reads `client.context["logdir"]` and gets the driver's basedir. And volara's consumer of it, `blockwise.py:284` (`set_log_basedir(client.context["logdir"])`), runs *inside* a spawned worker — the case daisy still covers.

**So this is migration work, not a merge blocker in principle**, but the tests must be updated before CI can go green: rewrite the two in-process assertions to exercise the spawn path (or mark them 1.x-only). The third failure, `test_a_spawned_process_recovers_the_drivers_basedir`, is the one that genuinely crosses the boundary and is worth keeping — it needs volara to export the context the way daisy's spawn path does.

Note this test file arrived from `main` via #47 and had never met daisy v2 before this merge — the branch had no `tests/test_logging.py` at all, so the conflict only surfaces here.

Separately, `Python Lint` is red for the one-line `measurements` import flagged in the suggestion above; that one is a clean fix.

volara/workers.py
```diff
@@ -37,139 +37,112 @@ def get_command(self, config_path: Path, task_name: str) -> list[str]:
         worker_id = context["worker_id"]
         task_id = context["task_id"]
 
-        worker_log_basename = daisy.get_worker_log_basename(worker_id, task_id)
+        worker_log_basename = daisy.logging.get_worker_log_basename(worker_id, task_id)
```
pattonw, line 41: this should be fixed in the daisy v1 compat layer

volara/workers.py
```diff
@@ -37,139 +37,112 @@ def get_command(self, config_path: Path, task_name: str) -> list[str]:
...
-        array_limit: int | None = None,
-        log_file: str | None = None,
-        error_file: str | None = None,
+        log_file: str | Path | None = None,
+        error_file: str | Path | None = None,
         flags: list[str] | None = None,
     ) -> list[str]:
         """
-        Prepares and optionally executes a command on a slurm cluster,
+        Build the ``srun`` line that runs one worker as a slurm job.
```
pattonw: switch to sbatch --wait instead of srun
See open PR for that

rhoadesScholar:
The `test_logging.py` port described in the diagnosis comment above is now open as #62 (against this branch): the three 1.x-form assertions rewritten to the v2 spawn-path contract, `4/4` on the ported file, full suite `186 passed, 4 skipped, 2 xfailed, 0 failed` at `f4497ec`. Once it merges into `will/rbws-plus-main`, the six red test legs here go green — leaving only the `extract_frags.py` F401, which already has the one-click suggestion above.

volara/datasets.py
```diff
@@ -202,14 +256,35 @@ def array(self, mode: OpenMode = "r") -> Array:
...
         self.lazy_ops(arr)
 
-        if self.channels is not None:
-            if isinstance(self.channels, list):
-                for channels in self.channels:
-                    arr.lazy_op(np.s_[channels,])
-            else:
-                arr.lazy_op(np.s_[self.channels])
+        ops = self.resolved_ops()
+        if mode != "r":
```
rhoadesScholar, line 260: @pattonw pretty sure there is a read only way to flip axes so this isn't an issue. I think that was my initial implementation...

PR #59, opened 2026-08-06
feat(workers)!: submit slurm workers with `sbatch --wait` instead of `srun`

rhoadesScholar:
> **For @pattonw to decide.** This changes how every slurm worker is submitted, and it removes the `srun` path rather than adding a mode beside it. Your call whether it belongs on this branch.

**BREAKING:** the `srun` submission path is removed. Slurm workers are now submitted with `sbatch --wait`.

> **Changed 2026-08-07, after @pattonw's review.** The teardown harness this PR originally carried is **gone** — see [What was removed](#what-was-removed-2026-08-07) at the bottom. This is now the submission change and nothing else.

## Why srun has to go

`srun` blocks, which is why it was chosen — but inside an existing allocation it creates a job **step**, not a new job. A driver that is itself a slurm job (the normal way to get a GPU node) therefore can never place workers anywhere but its own allocation, and `queue`/`--partition` is silently ignored for them.

Measured on a 32-CPU driver requesting 32 workers:

```
$ squeue -s -j 6677
6677.32   ...                       # ONE worker step
$ scontrol show step 6677.32
   Nodes=1 CPUs=32 Tasks=8          # 8 clones of one submission
$ grep -c "step creation still disabled" driver.log
623                                 # the other 31 workers, permanently blocked
```

Those 8 clones share a single `DAISY_CONTEXT`, hence one `worker_id` — the race daisy warns about in its own log line. Blocks still completed, so this degrades silently rather than failing. `num_workers` buys nothing.

That isn't a tuning problem; it's the whole fan-out. And it can't be fixed by launching differently — `sbatch` for the driver *is* the normal case.

## `sbatch --wait` satisfies the contract natively

The reason srun was picked over bare `sbatch` was *"no need for a custom wrapper around `subprocess.run()` with polling by process name."* `--wait` needs no wrapper either: it returns only when the job terminates, with the job's own exit status. Verified against a real scheduler:

```
$ time sbatch --wait --wrap="sleep 20; exit 7"
   → blocked 171s, returned rc=7
```

`blockwise.py` only requires the spawn command to block, so this is a drop-in — `subprocess.run(cmd)` is unchanged, and `blockwise.py`'s diff is now a **comment correction only**.

## The whole change

- `get_slurm_command` builds `sbatch --wait … --wrap=<command>` instead of `srun … <command>`. `shlex.join` for `--wrap`, since it is a shell string and a path with a space must survive as one argument.
- `--ntasks=1`, pinned explicitly. `sbatch` already defaults to one task, so this locks an invariant rather than fixing a live bug — but the failure it locks out is the measured one above, where a step took `ntasks` from the allocation.
- `is_srun_available` → `is_sbatch_available`.

Plus three things this PR falsified and had missed: the `tests/test_workers.py` module docstring still described `srun` as the fix, `tests/test_lambda_task.py` gated its real-cluster slurm test on `shutil.which("srun")`, and the `--ntasks` comment attributed the allocation-derived-ntasks behaviour to `sbatch` when it was measured for `srun` steps.

## The trade-off, stated plainly

`srun` tied the job to the client for free. An independent job does not, and **neither signal propagates** (measured on a real cluster, 25 s after each):

| action on the `sbatch --wait` client | worker job |
|---|---|
| `SIGKILL` | still `RUNNING` |
| `SIGTERM` | still `RUNNING` |
| `scancel <id>` | terminated |

So a driver that dies abnormally leaves its workers running until the partition's walltime, and there is **no knob here to bound that** — `flags` is not threaded through `SlurmWorker.get_command`, so a configured worker cannot request `--time` today. On the partitions measured here (`DefaultTime=NONE` / `MaxTime=UNLIMITED`) an escaped worker runs until reaped by hand with `scancel`.

This is recorded, not argued: it is documented in `get_slurm_command`'s docstring so the next reader finds it, and bounding it is a separate PR if wanted.

## Tests

5 worker tests (was 3): full-argv equality for the `sbatch --wait` line, `--ntasks=1` as a named regression carrying the measured failure mode, `shlex` quoting through `--wrap`, the LSF line, and `get_command` end to end.

```
tests/test_workers.py     5 passed
tests/                    183 passed, 4 skipped, 2 xfailed, 3 failed
```

The 3 failures are all in `tests/test_logging.py`, are **pre-existing** (reproduced on a clean clone of the base with identical errors), and are unchanged by rebuilding daisy to `b323d78` — I checked, since `KeyError: 'logdir'` looked like it might be a stale-daisy artifact. It isn't. `ruff check` and `ruff format` are both at exact base parity (1 pre-existing error, 4 pre-existing unformatted files).

## One thing worth knowing before this is useful

On our cluster this change is necessary but not sufficient: the installed daisy predated `7128d9c` *"Advertise an address workers can reach; stop binding loopback."* srun-as-steps put every worker on the driver's own node, where loopback resolves — so the step bug was **masking** the daisy bug. The moment workers land on other nodes they must be able to reach the driver. Anyone adopting this needs daisy ≥ `7128d9c`.

<a name="what-was-removed-2026-08-07"></a>

## What was removed 2026-08-07

@pattonw: *"I don't know if we need a large harness around the worker to make sure volara can kill it."* Removed, in `cf720a6`:

- the `Worker.run()` backend hook — dead code once `SlurmWorker.run` goes, no other backend overrode it
- `SlurmWorker.run()` — `Popen`, the job-id parse off sbatch's first stdout line, `scancel` in `finally`
- `blockwise.py` back to calling `subprocess.run(cmd)` directly
- `SlurmWorker.time_limit` and the `--time` flag, which existed **only** as the backstop for the one path teardown could not cover (a SIGKILLed driver runs no cleanup at all)
- the 5 teardown tests and the 2 `time_limit` tests

To answer the question directly, since it deserves an answer: the harness was never about a zombie *process* — the `sbatch --wait` client dies normally. It was about the slurm **job**, which survives its client; the table above is the measurement. So the MWE would have been "kill the driver, `squeue` still shows the workers." That is now an accepted cost rather than a mitigated one, and it is documented in the code.

https://claude.ai/code/session_014zvvAtcrDLNYu3uFBgw4Hm

### DIFF
--- a/tests/test_lambda_task.py
+++ b/tests/test_lambda_task.py
@@ -12,8 +12,8 @@ from volara.datasets import Labels, Raw
 from volara.workers import LocalWorker, LSFWorker, SlurmWorker
 
 slurm_available = pytest.mark.skipif(
-    shutil.which("srun") is None or os.environ.get("VOLARA_SLURM_TEST_QUEUE") is None,
-    reason="srun not found or VOLARA_SLURM_TEST_QUEUE not set",
+    shutil.which("sbatch") is None or os.environ.get("VOLARA_SLURM_TEST_QUEUE") is None,
+    reason="sbatch not found or VOLARA_SLURM_TEST_QUEUE not set",
 )
 lsf_available = pytest.mark.skipif(
     shutil.which("bsub") is None or os.environ.get("VOLARA_LSF_TEST_QUEUE") is None,
--- a/tests/test_workers.py
+++ b/tests/test_workers.py
@@ -26,8 +26,10 @@ import pytest
 from volara import workers
 
 _BODIES = {
-    # is_srun_available() probes `srun --version`.
-    "srun": 'echo "slurm-wlm 21.08.0"; exit 0',
+    # is_sbatch_available() probes `sbatch --version`.
+    "sbatch": 'echo "slurm-wlm 21.08.0"; exit 0',
+    # scancel is invoked by SlurmWorker.run's teardown.
+    "scancel": 'echo "$@" >> "$SCANCEL_LOG"; exit 0',
     # is_bsub_available() probes `bsub -V` (prints to stderr on real LSF).
     "bsub": 'echo "IBM Spectrum LSF 10.1"; exit 0',
 }
@@ -35,9 +37,11 @@ _BODIES = {
 
 @pytest.fixture()
 def cluster_shims(tmp_path, monkeypatch):
-    """Install fake srun/bsub on PATH so availability probes pass without a cluster."""
+    """Install fake sbatch/scancel/bsub on PATH so the availability probes and the
+    teardown run without a cluster. ``SCANCEL_LOG`` collects what teardown cancelled."""
     bindir = tmp_path / "cluster_shim_bin"
     bindir.mkdir()
+    monkeypatch.setenv("SCANCEL_LOG", str(tmp_path / "scancel.log"))
     for prog, body in _BODIES.items():
         script = bindir / prog
         script.write_text(f"#!/usr/bin/env bash\n{body}\n")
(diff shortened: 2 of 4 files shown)


pattonw:
I like the sbatch --wait change from srun. That seems fine. I don't know if we need a large harness around the worker to make sure volara can kill it.
Can you give me a minimum working example of a job that would create a zombie process that requires this harness to kill?

rhoadesScholar:
- [x] I'll pull out the wrapper and just leave the `sbatch --wait`

rhoadesScholar:
Done in cf720a6 — the harness is out, and this PR is now the submission change and nothing else. Title and body updated to match.

Removed: the `Worker.run()` backend hook, `SlurmWorker.run()` (Popen / job-id parse / `scancel` in `finally`), `SlurmWorker.time_limit` and the `--time` flag, and 7 tests. `blockwise.py` goes back to calling `subprocess.run(cmd)` directly, so its diff is now a **comment correction only**. What's left is `sbatch --wait`, `--ntasks=1`, `--wrap` with `shlex` quoting, and `is_sbatch_available` — 5 worker tests, `183 passed` with only the 3 pre-existing `test_logging.py` failures, ruff at exact base parity.

To answer your question, since it deserves an answer rather than being dropped: it wasn't a zombie *process*. The `sbatch --wait` client exits normally. It was the slurm **job**, which outlives its client — measured on a real cluster, 25 s after each signal:

| action on the `sbatch --wait` client | worker job |
|---|---|
| `SIGKILL` | still `RUNNING` |
| `SIGTERM` | still `RUNNING` |
| `scancel <id>` | terminated |

So the MWE would have been "kill the driver, then `squeue`" — the workers are still there. Under `srun` they weren't, because the step died with the client.

I've made that an accepted cost rather than a hidden one: `get_slurm_command`'s docstring now states it outright, including that `flags` isn't threaded through `SlurmWorker.get_command`, so a configured worker can't request `--time` today either. On our partitions (`DefaultTime=NONE` / `MaxTime=UNLIMITED`) that means an escaped worker runs until someone `scancel`s it. If you'd rather have a bound without the harness, a `time_limit` field alone is about ten lines and I'll open it as its own PR — say the word.

PR #60, opened 2026-08-06
fix(workers): resolve volara-cli from sys.executable, not the worker's PATH

rhoadesScholar:
Minimal, self-contained, and independent of the daisy-v2 line — this applies equally to `main` and to the migration branch.

## The bug

The worker command is built in the **driver** and executed on a **scheduler node**, in a shell the driver does not control. So a bare `volara-cli` resolves against the *worker's* `PATH` — and a scheduler starts the job in a fresh login shell, which commonly picks up a different environment entirely, typically a base conda env whose volara is stale and missing task entry-points.

The worker then fails pydantic's task-union validation and exits, and the driver waits forever for a worker that can never succeed. No crash, no progress — one observed instance orphaned 512 blocks with the driver looking healthy.

## The fix

```python
@staticmethod
def volara_cli() -> str:
    candidate = Path(sys.executable).parent / "volara-cli"
    return str(candidate) if candidate.is_file() else "volara-cli"
```

Resolving beside the running interpreter makes the worker environment independent of the job's `PATH` and conda activation. The fallback keeps unusual layouts working — zipapp, a PATH shim, an interpreter whose scripts live elsewhere — so behaviour is unchanged wherever there is no sibling script.

## Why it's worth doing upstream

Downstream (slabreg) has been carrying this as a `SlurmWorker` subclass that overrides `get_command` **purely to substitute `argv[0]`**. That subclass has broken twice:

- it once inlined a *copy* of `get_slurm_command`, which silently rotted when the signature changed and killed a full run;
- it drops any newly added kwarg, because it enumerates them by name — which just cost a walltime bound on a GPU fleet.

Neither failure would exist if the CLI resolved itself. With this merged, that subclass can be deleted.

## Tests

Adds `tests/test_workers.py` (new file on `main`): the pinned absolute path, the documented bare-name fallback, and the resulting argv.

```
tests/test_workers.py    3 passed
tests/                   177 passed, 4 skipped, 2 xfailed, 3 failed
```

The 3 failures are all `tests/test_logging.py` and are **pre-existing on `main`** — untouched by this PR, which changes only `argv[0]`.

### DIFF
--- a/tests/test_workers.py
+++ b/tests/test_workers.py
@@ -1,0 +1,45 @@
+"""Worker command construction.
+
+The worker command is built in the DRIVER but executed on a scheduler node, in a shell the
+driver does not control -- so anything it resolves by name is resolved against the worker's
+environment, not the driver's.
+"""
+
+import sys
+from pathlib import Path
+
+from volara import workers
+
+
+# ------------------------------------------------- volara-cli resolution (PATH-proof) ---
+def test_worker_command_pins_the_cli_beside_this_interpreter(monkeypatch):
+    """A bare ``volara-cli`` resolves against the WORKER's PATH, not the driver's. A
+    scheduler starts the job in a fresh login shell, so it can pick up a different
+    environment -- typically a base conda env with a stale volara missing task
+    entry-points -- and the worker then fails task-union validation while the driver
+    waits forever for a worker that can never succeed."""
+    # LocalWorker.get_command reads the daisy worker context; supply one so this test
+    # needs no running daisy server.
+    monkeypatch.setenv("DAISY_CONTEXT", "worker_id=0:task_id=t")
+    expected = str(Path(sys.executable).parent / "volara-cli")
+    cmd = workers.LocalWorker().get_command(Path("/tmp/c.json"), "t")
+    if Path(expected).is_file():
+        assert cmd[0] == expected, cmd
+    else:  # no sibling CLI in this environment -> documented PATH fallback
+        assert cmd[0] == "volara-cli", cmd
+    assert cmd[1:] == ["blockwise-worker", "-c", "/tmp/c.json"], cmd
+
+
+def test_cli_falls_back_to_the_bare_name_when_there_is_no_sibling(monkeypatch, tmp_path):
+    """Unusual layouts (zipapp, shim on PATH, scripts installed elsewhere) must keep
+    working exactly as before rather than pointing at a path that does not exist."""
+    monkeypatch.setattr(sys, "executable", str(tmp_path / "nonesuch" / "python"))
+    assert workers.Worker.volara_cli() == "volara-cli"
+
+
+def test_cli_is_absolute_when_a_sibling_exists(monkeypatch, tmp_path):
+    fake_bin = tmp_path / "bin"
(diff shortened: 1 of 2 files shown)


pattonw:
I've seen the subclass in `slabreg` and come across this explanation before. I'm not convinced.

This needs more justification:
> So a bare volara-cli resolves against the worker's PATH — and a scheduler starts the job in a fresh login shell, which commonly picks up a different environment entirely, typically a base conda env whose volara is stale and missing task entry-points.

Especially the "commonly picks up a different environment" part. We've been using volara, slurm, lsf, and other clusters for years. Usually when you submit a job on the cluster, you can activate your environment, call `srun ... my_script.py` and expect the job to run on the same environment as you are currently on because environment variables get passed through to the job. That is the same design pattern we are depending on here. We are expecting users to have a clean environment, with only one version of `volara` installed, so calling the command line entry point to volara should be called within the same environment as the server.


My concerns with this approach is mostly just that it might hide a messy environment, leading to more confusing issues for people down the line. With `uv` it is now very fast/easy to strictly define and rebuild environments regularly and on demand, so I think its fine to assume a clean environment and not build in extra safeguards.

Open to changing my mind if there are more common reasons why a different environment might be picked up.

rhoadesScholar:
You're right, and I measured it rather than argue — thanks for pushing back.

Your premise was `srun`, and #59 has since replaced that with `sbatch --wait`, so it needed re-testing either way. On the AWS cluster, with `sbatch --wait --wrap` (what volara does now):

**Simple case** — submit from an activated env:
```
SUBMITTER  which volara-cli : /home/jeffr/miniforge3/envs/slabreg/bin/volara-cli
WORKER     which volara-cli : /home/jeffr/miniforge3/envs/slabreg/bin/volara-cli
WORKER     CONDA_DEFAULT_ENV: slabreg
```

**Production topology** — an sbatch *driver* that sets `PATH` by `export` rather than `conda activate` (what our driver script actually does), which then submits an sbatch worker. Two hops, the harder case:
```
DRIVER  which volara-cli : /home/jeffr/miniforge3/envs/slabreg/bin/volara-cli
WORKER  which volara-cli : /home/jeffr/miniforge3/envs/slabreg/bin/volara-cli
WORKER  CONDA_DEFAULT_ENV: slabreg
WORKER  PATH head        : /home/jeffr/miniforge3/envs/slabreg/bin:...
```

`sbatch`'s default `--export=ALL` carries `PATH` through both hops; `python` and the `volara` module path matched the submitter's exactly, including the editable checkout. **The premise doesn't hold, so closing this.**

Your second point is the one that should outlive the PR: a safeguard like this would have *hidden* a messy environment rather than surfaced it. I'm also deleting slabreg's `VenvSlurmWorker`, which existed for exactly this reason — and which independently re-enumerates `get_slurm_command`'s kwargs by name, so it silently drops any field volara adds (it did exactly that once, costing a full-volume run).

One caveat for the record, not an argument to keep it: this is one cluster with default `--export=ALL`. A site configured `--export=NONE`, or with a prolog that resets `PATH`, would behave differently — and if that ever comes up, the fix belongs here and conditional, not as a permanent downstream subclass.

PR #61, opened 2026-08-07
Add s3 uri option for lut class

sheridana:
Enables s3 storage of luts, e.g:

```py
lut_path = "s3://bucket/path/to/lut.npz"
lut = LUT(path=lut_path)
```

### DIFF
--- a/tests/test_lut.py
+++ b/tests/test_lut.py
@@ -1,6 +1,23 @@
+import fsspec
 import numpy as np
+import pytest
 
 from volara.lut import LUT, LUTS
+
+
+@pytest.fixture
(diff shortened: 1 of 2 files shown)


pattonw:
Merging for now, since this seems like a useful feature that is holding us back. But we should consolidate our interface with files and s3 links. We have at least 4 cases that we want to handle regularly:
- files (original implementations and supported almost everywhere)
- s3 buckets that can be easily accessed via configured profiles (supported most places)
- s3 buckets that need kwargs: lyvecloud (supported most places s3 is supported, not sure if this is tested quite as well)
- google buckets (not sure if this is supported anywhere)

PR #62, opened 2026-08-08
test(logging): pin the daisy-v2 spawn-path logdir contract instead of the 1.x every-Context form

rhoadesScholar:
Fixes the three `tests/test_logging.py` failures that keep all six of #58's CI test legs red ([diagnosis comment](https://github.com/e11bio/volara/pull/58#issuecomment-5198769959)). Test-only change; no library code is touched.

### Why the tests were red

The tests asserted daisy **1.x**'s contract: every `Context` carries `logdir` at construction. daisy **2.0** (this branch's pin) no longer offers that contract anywhere:

- a bare `daisy.Context(...)` is the Rust `PyContext` and injects nothing (`daisy-py/src/py_context.rs:27-39` at `v2.0` tip `b323d78`);
- `logdir` is injected only when `DAISY_CONTEXT` is built for a **spawned worker**: `context_with_logdir()` at `daisy-py/python/daisy/_worker_processes.py:118-144`, applied at `:247`, reading the driver's `daisy.logging` global — which `volara.logging.set_log_basedir` forwards to (`volara/logging.py:25`).

So the failures were by design, not a regression: `KeyError: 'logdir'` on a context daisy never claims to decorate.

The contract volara actually relies on still holds. The worker-side consumer, `blockwise.py:284` (`set_log_basedir(client.context["logdir"])`), runs inside a spawned worker — exactly the covered path — and daisy's own `tests/test_worker_execution.py:232-265` and `:269-310` assert a spawned worker (and even a re-exec'ed grandchild) recovers the driver's basedir. End-to-end, this suite already exercises it under a custom basedir on every multiprocessing test, because `tests/conftest.py`'s autouse `logdir` fixture points each test at a private basedir — a worker falling back to the default would address a different `blocks_done` store and fail those tests.

### The port

Each in-process test now pins the **spawn-path** form via a `_spawn_context()` helper; the first test also asserts that a bare `Context` carries **no** `logdir`, so the file cannot drift back to the 1.x form. `test_a_spawned_process_recovers_the_drivers_basedir` keeps its real-child-process check, with `DAISY_CONTEXT` holding what daisy's spawn path puts there. `test_a_worker_without_a_context_does_not_silently_guess` is unchanged.

`_spawn_context()` deliberately imports daisy's private `daisy._worker_processes.context_with_logdir`: volara's worker reads what that function wrote, so if daisy renames or reshapes it, volara should find out from this file rather than from a full-volume run (rationale in the helper's docstring).

### Reproduction

Env: `uv sync --extra dev --extra indirect`, Python 3.13.13, daisy 2.0.0 resolved from `v2.0` @ `b323d78` (contains funkelab/daisy@505e7be).

**On the base branch (`will/rbws-plus-main` @ `8a78749`):**
```console
$ uv run pytest tests/test_logging.py -q
FAILED tests/test_logging.py::test_volara_set_log_basedir_reaches_the_daisy_context
  - KeyError: 'logdir'   # Path(_context()["logdir"])
FAILED tests/test_logging.py::test_the_basedir_survives_serialization_to_the_environment
  - AssertionError: assert 'logdir' in 'task_id=t:worker_id=0:port=1234:hostname=localhost'
FAILED tests/test_logging.py::test_a_spawned_process_recovers_the_drivers_basedir
  - CalledProcessError (child: KeyError: 'logdir')
3 failed, 1 passed
```

**On this branch:**
```console
$ uv run pytest tests/test_logging.py -q
4 passed
```

**Full suite:** base `3 failed, 183 passed, 4 skipped, 2 xfailed` → this branch **`186 passed, 4 skipped, 2 xfailed`**.

### Not included

The `Python Lint` leg's `extract_frags.py` F401 already has a one-click suggestion on #58 — separate change, deliberately not folded in.

https://claude.ai/code/session_017QGnabZYYcDs5NLMtnMq8t

### DIFF
--- a/tests/test_logging.py
+++ b/tests/test_logging.py
@@ -2,7 +2,7 @@
 
 volara spawns each worker as a fresh ``volara-cli`` PROCESS, so nothing that lives only in
 a module-global reaches it.  ``BlockwiseTask.process_blocks`` recovers the driver's basedir
-from ``daisy.Client().context["logdir"]`` (``volara/blockwise/blockwise.py:260``) -- i.e.
+from ``daisy.Client().context["logdir"]`` (``volara/blockwise/blockwise.py:284``) -- i.e.
 volara depends on **daisy** to carry it across the boundary, in the ``DAISY_CONTEXT``
 environment variable.
 
@@ -12,13 +12,20 @@ worker that disagree about it address *different* done-marker stores, the worker
 ``open_ds(..., mode="a")`` creates a zarr Group where an Array is expected, and the run
 orphans **every** block while the driver log stays clean.
 
-It has already broken once: daisy v2 dropped ``logdir`` from the worker context and replaced
-it with a fallback that reads the process-global basedir -- correct for a *forked* worker,
-wrong for a spawned one.  Fixed upstream in funkelab/daisy@505e7be.
+WHERE daisy carries it moved in daisy 2.0, and these tests pin the new place.  daisy 1.x
+put ``logdir`` on every ``Context`` at construction; daisy 2.0 injects it only when
+``DAISY_CONTEXT`` is built for a spawned worker (``context_with_logdir()`` in
+``daisy/_worker_processes.py``), reading the driver's ``daisy.logging`` global -- which
+``volara.logging.set_log_basedir`` forwards to.  A bare ``daisy.Context(...)`` never
+carries ``logdir`` under v2, so asserting the 1.x form means asserting a contract daisy
+no longer offers anywhere.  What volara actually relies on is the spawn path, and that is
+what is pinned here.
 
-There was no test in volara crossing a process boundary at all, which is why a change to
-someone else's contract could do that silently.  These tests pin the contract volara
-actually relies on, so the next change to it fails here instead of in a full-volume run.
+The end-to-end form of the same contract -- a real spawned ``volara-cli`` worker marking
+blocks done in the store the driver prepared -- is already exercised by every
+multiprocessing test in this suite: ``tests/conftest.py``'s autouse ``logdir`` fixture
+points each test at a private basedir, so a worker that fell back to the default would
+address a different ``blocks_done`` store and fail those tests loudly.
 """
 
 import json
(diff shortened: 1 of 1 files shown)


pattonw:
This seems like a daisy v2 bug. We don't want to break any behavior that users depend on when upgrading to v2. If we're expecting logdir to be available somewhere and it isn't then we should probably fix that.
It seems like the logdir is being set based on the environment variables, and if created in the server then that variable is not available so daisy doesn't set it.

We probably also don't want to be depending on hidden modules in daisy. I.e. the `context_with_logdir` import.

rhoadesScholar:
Both points taken — and you're right that it's daisy's bug, not just a test mismatch. The daisy side is now up as a draft: https://github.com/funkelab/daisy/pull/79 (draft).

What convinced me it's a genuine v2 regression rather than volara over-asserting: the spawn boundary has two *consumer-visible* channels — the process-global `DAISY_CONTEXT` env var (whose own code comment promises "the canonical cluster-worker pattern" keeps working) and the keyword-only `context` argument handed to spawn functions — and both emit five-key, logdir-less contexts; only daisy's private subprocess wrapper patches the key in. Worse, absence is silent: `Client` falls back to the worker's *own* default when `logdir` is missing (`_task.py:336-341`), which is exactly the silent-guess failure mode this test file's fourth test exists to forbid (a worker on the wrong basedir resolves a different `blocks_done` store and orphans every block with a clean driver log).

The daisy draft completes the context once in `PySpawnWorker::spawn` — the last point both channels share — so every channel emits the full context, and it makes `context_with_logdir` public (`daisy.logging.context_with_logdir`, re-exported at top level like `get_worker_log_basename`), with the private name kept as an alias. Evidence in its description: the two new daisy tests fail on `v2.0`@`b323d78` (`KeyError: 'logdir'` on the env-var channel; no public name) and pass on the branch; the full daisy suite runs at exact stock parity (241 passed — the only 2 failures are hardware-parallelism tests that fail identically on unpatched `b323d78` on the same 2-vCPU node).

For this PR, your call on sequencing:

1. **Merge this now, swap the import after daisy lands** — the `_spawn_context()` helper's private import becomes `daisy.context_with_logdir` (a 2-line follow-up), and #58's six red test legs go green today; or
2. **Hold this until the daisy fix is on `v2.0`**, and I rebase it onto the public name so no private import ever lands.

Either way the spawn-path test shape stays — a bare driver-side `Context()` deliberately stays `logdir`-less under v2's design (the daisy draft restores the *channels*, not construction-time injection), so the three rewritten tests here remain the correct form.

PR #63, opened 2026-08-08
feat(workers): let a SlurmWorker ask for memory

rhoadesScholar:
`get_slurm_command` has always accepted a `memory=` argument. `get_command` — the path daisy
actually spawns workers through — never passed one. So **every** `SlurmWorker` silently got the
15564 MB default, and there was no way to ask for more.

## Why this is a ceiling, not a tuning nicety

It caps what a blockwise task can do *at all*. A task whose block is large cannot run on a worker,
and dispatching such a task to the fan-out path moves the computation to a machine **smaller than
the driver it came from**.

The case that surfaced it: `volara-surface`'s surface postprocess has a whole-face `write_roi`, so
`VOLARA_SURFACE_DISPATCH=blockwise` decomposes it into exactly one block. On a 24414×25016 face that
block needed **~134 GiB**. It was dispatched to a **15.2 G** worker. Forcing blockwise looked like
the fix for a driver OOM and could never have worked — the same computation, on 1/8th the machine.

## The change

One field on `SlurmWorker`, threaded through `get_command`. **Default unchanged**, so no existing
pipeline moves.

```python
class SlurmWorker(Worker):
    queue: str
    num_gpus: int = 0
    num_cpus: int = 1
    memory: int = 15564   # new
```

## Gates

| test | guards |
|---|---|
| `test_get_command_requests_the_workers_configured_memory` | a configured `memory` reaches `--mem` — **verified to FAIL when the passthrough line is removed**, which is the whole change |
| `test_worker_memory_defaults_to_the_historical_value` | an unconfigured worker still asks for exactly 15564, so nobody's pipeline shifts under them |

7 tests in `test_workers.py` pass.

## Scope note

**Not touching `LSFWorker`.** `get_lsf_command` takes no memory argument at all, so the equivalent
there means changing the command builder itself, and I have no LSF cluster to test against. Flagging
rather than guessing.

Based on `will/rbws-plus-main` (where #59 landed) rather than `main`, since that is the v2 line this
pipeline runs.

### DIFF
--- a/tests/test_workers.py
+++ b/tests/test_workers.py
@@ -152,3 +152,45 @@ def test_get_command_names_job_after_task(cluster_shims, monkeypatch, tmp_path):
     cmd = workers.SlurmWorker(queue="gpu-q").get_command(tmp_path / "c.json", "mytask")
     assert cmd[0] == "sbatch" and cmd[1] == "--wait", cmd
     assert "--job-name=mytask" in cmd, cmd
+
+
+def test_get_command_requests_the_workers_configured_memory(
+    cluster_shims, monkeypatch, tmp_path
+):
+    """REGRESSION. `get_slurm_command` has always accepted `memory=`, but `get_command` -- the path
+    daisy actually spawns workers through -- never passed one, so EVERY SlurmWorker silently got the
+    15564 MB default with no way to ask for more.
+
+    That is not a tuning nicety: it caps what a blockwise task can do at all. A task whose block is
+    large cannot run on a worker, and pushing such a task onto the fan-out path moves the
+    computation to a machine SMALLER than the driver it came from. e11bio/volara-surface's
+    whole-face surface postprocess hit exactly this -- a 134 GiB computation dispatched to a 15.2 G
+    worker."""
+    import daisy
+
+    monkeypatch.setenv("DAISY_CONTEXT", "worker_id=0:task_id=mytask")
+    monkeypatch.setattr(
+        daisy.logging, "get_worker_log_basename", lambda wid, tid: tmp_path
+    )
+
+    cmd = workers.SlurmWorker(queue="q", memory=131072).get_command(
+        tmp_path / "c.json", "mytask"
+    )
+    assert "--mem=131072" in cmd, cmd
+    assert "--mem=15564" not in cmd, cmd
+
+
+def test_worker_memory_defaults_to_the_historical_value(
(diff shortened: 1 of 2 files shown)


rhoadesScholar:
**CI triage — the test failures are the base branch's own baseline, not this change.**

Measured rather than asserted. Same suite, same interpreter, two commits:

| commit | `tests/test_logging.py` |
|---|---|
| `8a78749` — untouched `will/rbws-plus-main` (the base) | **3 failed**, 1 passed |
| this branch | **the same 3 failed**, and 185 passed / 4 skipped / 2 xfailed overall |

The three are `test_volara_set_log_basedir_reaches_the_daisy_context` (`KeyError: 'logdir'`),
`test_the_basedir_survives_serialization_to_the_environment`, and
`test_a_spawned_process_recovers_the_drivers_basedir`. All three are about daisy's logdir/context
plumbing; none of them touches `workers.py`, and this PR adds one pydantic field plus one keyword
argument.

`ty` and `Python Lint` are likewise red on the base.

So the delta from this branch is **+2 tests, both passing**, and no change to any existing result.
Happy to rebase if the base goes green first, but nothing here is waiting on me.

rhoadesScholar:
Converted to draft to comply with the standing rule (2026-08-08): PRs against `daisy`, `funlib.*` and `volara` open as drafts and stay drafts until Jeff approves marking them ready. This one predates the rule by about an hour; no change to the content.

Substance is unchanged and it is complete as far as I can take it — one field, default preserved, two gates (one verified to fail without the passthrough), and the CI triage above showing the red checks are the base branch's own.

pattonw:
This is a solution for a job that broke a blockwise task down into a single block and OOM'd? That seems like a logic bug that should be fixed so that a task has an appropriate block size.

Is the main use case that we could use more generic queues that would distribute workers across multiple different hardware queues with different capacities, and as long as it is more than the amount set we'll be fine? So far we've solved this by just requesting the queue we need based on the requirements of each job. Is that insufficient for your use case?

rhoadesScholar:
> This is a solution for a job that broke a blockwise task down into a single block and OOM'd? That seems like a logic bug that should be fixed so that a task has an appropriate block size.

Doesn't MWS end with a single full FOV block? We haven't had to worry about it being too big to do the final agglomeration... yet.

That said, totally fine if you don't want to add this. 👍🏼

PR #65, opened 2026-08-12
fix: blockwise tasks never resume — check_function always raises on daisy v2

rhoadesScholar:
`check_function` raises on every block, so no block is ever seen as done and every rerun recomputes everything.

### Reproduction

One daisy task, 64 blocks, run twice. Second run should skip all 64.

```python
def check_block(block):                                  # volara/blockwise/blockwise.py:204-215
    block_array = open_ds(block_ds_path, mode="r")
    coordinate = (block.write_roi.offset - block_array.offset) / block_array.voxel_size
    ...

task = daisy.Task(..., check_function=check_block)
daisy.run_blockwise([task], multiprocessing=False)       # twice
```

**On `main`:**
```
run 1: process_function called on 64 blocks
run 2: process_function called on 64 blocks
check received block.write_roi.offset of type: daisy._daisy.Coordinate
check RAISED: TypeError: unsupported operand type(s) for -: 'daisy._daisy.Coordinate' and 'Coordinate'
```

**On this branch:**
```
run 1: process_function called on 64 blocks
run 2: process_function called on 0 blocks
Execution Summary:  blocks 64   completed 64   skipped 64
```

### Cause

daisy v2's compat layer wraps `process_function` so it receives a funlib-typed `Block`
(`daisy/v1_compat.py:221-225`) but does not wrap `check_function`. `check_block` therefore gets a
native `Block`, and `blockwise.py:209` evaluates `daisy._daisy.Coordinate - funlib.geometry.Coordinate`:

```python
>>> daisy.Coordinate((0,0,0)) - funlib.geometry.Coordinate((1,1,1))
TypeError: unsupported operand type(s) for -: 'daisy._daisy.Coordinate' and 'Coordinate'
```

The scheduler catches a raising check as "not done", so the failure is silent.

### Fix

Use daisy's built-in done-markers rather than checking `blocks_done.zarr` through `check_function`.
`tracking_path` is under `meta_dir`, so `drop()` still clears the done state and rerun semantics are
unchanged — markers outside `meta_dir` would survive a drop and skip everything.

`blocks_done.zarr` is left in place; removing it is a separate change.

### Tests

`3 failed, 182 passed` on both `main` and this branch — the three `test_logging` failures are
pre-existing and unrelated.

Related: funkelab/daisy#80 wraps `check_function` upstream, which fixes released volara versions too.
These are complementary.

### DIFF
--- a/tests/test_lut.py
+++ b/tests/test_lut.py
@@ -1,6 +1,23 @@
+import fsspec
 import numpy as np
+import pytest
 
 from volara.lut import LUT, LUTS
+
+
+@pytest.fixture
+def fake_s3(monkeypatch):
+    """
+    Back `LUT._s3fs` with an in-memory fsspec filesystem so that s3 code
+    paths can be exercised without a bucket.
+    """
+    fs = fsspec.filesystem("memory")
+    fs.store.clear()
+    fs.pseudo_dirs.clear()
+    monkeypatch.setattr(LUT, "_s3fs", lambda self: fs)
+    yield fs
+    fs.store.clear()
+    fs.pseudo_dirs.clear()
 
 
 def test_lut_save_load(tmp_path):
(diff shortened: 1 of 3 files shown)


rhoadesScholar:
**Review, rewritten 2026-08-12** — the base blocker is resolved and the conflict is fixed, so this replaces a much longer earlier version.

**Conflict fixed** (`270c31b`). Merge, not rebase, so nothing was force-pushed. The base renamed `num_workers` → `max_workers`; this PR swaps `check_function` for `tracking_path`. Resolved as the union of both — `max_workers` matters because `num_workers` only survives on a daisy-1.x compat alias that warns.

**MWE** — one task, run twice, counting `process_block` calls:

```
base    run 1: 64 blocks   run 2: 64 blocks   ← silently recomputes everything
branch  run 1: 64 blocks   run 2:  0 blocks   ← resumes
```

**Tests:** 191 passed, 3 failed. The same 3 (`tests/test_logging.py`) fail on the base with this branch absent — pre-existing, not from this change.

**Two things before it leaves draft:**

1. **Add the regression test** — run twice, assert zero recomputation, and assert a *failed* block is still retried. For a bug whose defining property was silence, no test is the wrong economy.
2. **Fix the Tests section** — "3 failed, 182 passed on both `main` and this branch" was measured before the retarget. Use the numbers above.

📌 `check_block_func` now has no production caller (this PR removed its only one) but still has a test. Left alone deliberately: removing it is a different change from fixing resume.

pattonw:
This is the correct fix, but includes some changes on LUTs, and doesn't purge volara of its own custom block done handling. If we rely on daisy's built in version, we no longer need our own check_block and mark_block_done functions, or our own block_done dataset. I merged a similar fix that also purges the old implementation

PR #66, opened 2026-08-21
feat(datasets): Dataset.flip reverses axes of the final array

rhoadesScholar:
Adds `Dataset.flip: list[int] | None` — axis indices to reverse — so a volume acquired in a flipped orientation can be corrected by declaration rather than by materialising a corrected copy.

Motivation is slabreg: a gel slab is sometimes imaged upside down, and today that is handled by producing a duplicate volume outside the pipeline. That costs a full copy of the data and leaves the correction invisible in the config that describes the dataset.

### Why a field, and not a wrapper or a subclass

A blockwise worker rebuilds the task model from `model_dump_json()` and calls `array()` itself. A **declared field travels**; an in-memory wrapper does not. Neither does a `Dataset` subclass — assigned to a `Raw`-typed task field the subclass survives in memory, but `model_dump()` drops the added field and the round-trip returns a plain `Raw`:

```
direct dump keeps flip:                     True
after assignment to a `Raw`-typed field:    RawFlipped   # still the subclass, in memory
serialised keys include flip:               False        # <-- dropped here
round-tripped type:  Raw  | flip survives:  GONE
```

That applies in the driver and **silently not under slurm**, which is the worst possible split — it passes exactly the smoke test you would run first. Two tests pin the travelling property directly, one through a `Raw`-typed task field.

`channels` is the existence proof that this shape works: it is already a declared field whose lazy op survives the hop.

### Three deliberate choices

- **Applied after the `channels` collapse**, not in `lazy_ops` — `lazy_ops` runs at `:203`, *before* the channels block at `:205-213`, so a flip there would land on the timepoint/channel axes of a 5-D store. The indices name the **final** array: `[T,C,Z,Y,X]` read with `channels=[0,0]` is `[Z,Y,X]`, and `flip=[0]` reverses Z. There is a test for exactly this.
- **A slice tuple, not a callable**, so funlib keeps its introspectable fusion path (and this does not copy the late-binding closure over the loop variable at `:209`).
- **On `Dataset`, not `Raw`**, so `Dataset.spoof()` carries it for `BlockwiseTask.benchmark()`.

### Read-only, and one companion fix

A reversed array reports `is_writeable` True while `__setitem__` raises, so a non-`"r"` mode is refused up front rather than at the first write.

`CloudVolumeWrapper.array()` fully overrides `Dataset.array()` and applies neither `lazy_ops` nor `channels`, so a base-class `flip` would be **silently inert** there — the exact failure this design exists to avoid, relocated. It now refuses `flip` and `channels` by name. That also surfaces a pre-existing silent ignore of `channels`.

### Verification

10 new tests: single and multi-axis reversal; indices resolve against the post-`channels` array; **sub-ROI reads are exact** (a worker reads blocks, so whole-array correctness is not enough); round-trip through JSON and through a task-typed field; `flip=None`/`[]` byte-identical to before; out-of-range axis raises naming the final ndim; write mode refused; CloudVolume refuses.

The existing suite is **25 failed / 164 passed both with and without this change** on this branch — a daisy API mismatch against the checkout's venv (`tracking_path`, `init_block_array`), pre-existing and unrelated. Verified by exporting the unmodified base with `git archive` and running both.

Targets `will/rbws-plus-main` because that is the branch slabreg pins (`uv.lock`: `?branch=will%2Frbws-plus-main`), not `main`.

https://claude.ai/code/session_01FbGpCGKwenjmLG6quBEw6Y

---

### ⚠️ CI state on this branch, before you read the red marks

**`will/rbws-plus-main` is already failing CI on its own head** — `gh run list --branch will/rbws-plus-main` shows `Python Ruff: failure` and `ty: failure`. This PR inherits that; it does not cause it.

- **Ruff** — was one unused `scipy.ndimage.measurements` import in `extract_frags.py`, a file this PR does not otherwise touch. Fixed here in a separate commit, because leaving the PR red for a reason that is not its own is worse than a one-line drive-by. Ruff now passes.
- **`ty` and the `test-*` matrix** — pre-existing and not mine. The test failures are a daisy API mismatch (`'DummyTask' object has no attribute 'init_block_array'`, `KeyError: 'logdir'`, `Task.__new__() got an unexpected keyword argument 'tracking_path'`), which is what this branch's daisy pin does against the current suite.

**Evidence that this change adds nothing:** I exported the unmodified base with `git archive` (not `cp -a` — a worktree's `.git` is a pointer file and a copy writes through to the real branch) and ran both trees against the same interpreter:

```
base (origin/will/rbws-plus-main, unmodified):  25 failed, 164 passed, 4 skipped, 1 xfailed
this branch, excluding its own new tests:       25 failed, 164 passed, 4 skipped, 1 xfailed
this branch's new tests:                        10 passed
```

Identical counts, same named failures. `ruff check` on the two files this PR authors: clean.

### DIFF
--- a/tests/test_dataset_flip.py
+++ b/tests/test_dataset_flip.py
@@ -1,0 +1,111 @@
+"""`Dataset.flip` reverses axes of the FINAL array, and travels in the config.
+
+The point of a declared field rather than a wrapper: a blockwise worker rebuilds the model from
+`model_dump_json()` and calls `array()` itself, so the reversal applies there too. A subclass or an
+in-memory wrapper applies in the driver and silently not in the worker.
+"""
+
+import numpy as np
+import pytest
+from funlib.geometry import Coordinate, Roi
+from funlib.persistence import prepare_ds
+
+from volara.datasets import Raw
+
+
+def _store(tmp_path, shape, name="t.zarr"):
+    p = tmp_path / name
+    arr = prepare_ds(
+        p, shape=shape, dtype=np.uint16, voxel_size=(1,) * len(shape),
+        offset=(0,) * len(shape), axis_names=list("tczyx"[-len(shape):]),
+        units=["nm"] * len(shape), mode="w",
+    )
+    data = np.arange(int(np.prod(shape)), dtype=np.uint16).reshape(shape)
+    arr[:] = data
+    return p, data
+
+
+def test_flip_reverses_the_named_axis(tmp_path):
+    p, data = _store(tmp_path, (6, 5, 4))
+    r = Raw(store=p, flip=[0], voxel_size=(1, 1, 1), offset=(0, 0, 0))
+    assert np.array_equal(np.asarray(r.array("r")[:]), data[::-1])
+
+
+def test_flip_takes_several_axes(tmp_path):
+    p, data = _store(tmp_path, (6, 5, 4))
+    r = Raw(store=p, flip=[0, 2], voxel_size=(1, 1, 1), offset=(0, 0, 0))
+    assert np.array_equal(np.asarray(r.array("r")[:]), data[::-1, :, ::-1])
+
+
+def test_flip_indexes_the_array_AFTER_channels(tmp_path):
+    """The whole ndim trap: on a 5-D store `z` is axis 2, but axis 0 once channels collapse."""
+    p, data = _store(tmp_path, (2, 3, 6, 5, 4))
+    r = Raw(store=p, channels=[1, 2], flip=[0], voxel_size=(1,) * 5, offset=(0,) * 5)
+    got = np.asarray(r.array("r")[:])
+    assert got.shape == (6, 5, 4)
+    assert np.array_equal(got, data[1, 2][::-1])
+
+
+def test_sub_roi_reads_through_the_reversal_are_exact(tmp_path):
+    """A blockwise worker reads BLOCKS, so whole-array correctness is not enough."""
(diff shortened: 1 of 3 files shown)


rhoadesScholar:
Follow-up opened: **#67 — lazy ops become an ordered list, keywords deprecated**, stacked on this branch.

It came out of reviewing this PR's weakest point. `flip` here is applied at the *end* of `array()`, after the `channels` collapse, and the axis indices are only correct because of that placement — but nothing states or enforces it. Put the same op in `lazy_ops()` instead (the documented extension hook, and the natural place for it) and it runs *before* `channels`, so a 4-D read silently returns **the wrong channel, unflipped**. Same field, same value, no error.

That is a property of the whole op set, not of `flip`: `ome_norm` has the opposite constraint — it indexes the channel axis, so it must run *before* the collapse. The ordering is load-bearing and currently implied by the order of `if` statements.

#67 makes it explicit as `ops=[...]`, keeps the keyword form byte-identical behind a `DeprecationWarning`, and refuses the ambiguous both-set case. The one design point worth flagging for review: the ops are **declarative models, not callables**. A `list[Callable]` cannot `model_dump_json()`, and `blockwise.py:198` writes exactly that to the `config.json` a worker reads — so callables would work in-process and break every distributed run. There's a test asserting it, because it is the reason for the API's shape.

Reviewing them together may be easier than #66 alone, since #67 is what makes this PR's ordering requirement checkable rather than assumed.

PR #67, opened 2026-08-22
feat(datasets): one place for the lazy-op order, plus a trailing callables hook

rhoadesScholar:
Implements #30. Stacked on #66 — base is `flip/dataset-axis-reversal`, 0 commits behind it.

Puts the order of a `Dataset`'s lazy operations in **one place**, and adds a trailing hook for transforms that have no keyword.

### Why

Ops are applied in an order fixed by the order of `if` statements in `Dataset.array()`, and that order is **load-bearing, with two constraints that bind in opposite directions**:

- `ome_norm` indexes the channel axis, so it must run **before** `channels` collapses it.
- `flip` (#66) names the collapsed array, so it must run **after**.

Nothing states this. A reader can't see it, and it isn't checkable. That is not hypothetical: `lazy_ops()` is the documented extension hook and the natural place to land a new op — and landing `flip` there instead of at the end makes a 4-D read silently return **the wrong channel, unflipped**. Same field, same value, no error.

### What changes

New `volara/ops.py` — one model per keyword (`OmeNormalize`, `ScaleShift`, `StackWith`, `SelectChannels`, `ReverseAxes`). `Dataset.resolved_ops()` returns them in the one correct order, `Raw` prepending its three to the base class's two, and `array()` applies that list:

```python
Raw(store=…, ome_norm=meta, channels=[0, 0], flip=[0], ops=[lambda d: d * 2])
# resolved_ops() -> OmeNormalize, SelectChannels, ReverseAxes, then the callable
```

New `Dataset.ops: list[PydanticCallable] | None` — plain callables applied after every keyword op, for a transform that has no keyword. They ship cloudpickled to base64, exactly as `LambdaTask.lambda_func` already does, so they cross into a worker.

Models rather than inlined `if`s buys two things: the order becomes **assertable**, and the ops stay **introspectable** — a consumer can read the reversed axes off a `ReverseAxes`, where a cloudpickled callable exposes nothing.

`Raw.lazy_ops` is removed. `Dataset.lazy_ops` stays as the subclass hook, still called first.

### Points for review

**One behaviour change against the base branch, and it reaches no plain `Raw`.** Removing `Raw.lazy_ops` moves its three keyword ops relative to a *subclass* that overrides the hook: after the subclass's own ops if it calls `super()`, and applied at all if it doesn't — before, an override that skipped `super()` silently disabled normalization. Position relative to `channels`/`flip` is unchanged. A `grep` over volara, volara-registration, volara-surface, slabreg and e11-models finds no such subclass.

**Three refusals, all up front rather than at first use:**

- `ops` in a write mode — a callable makes the funlib array unwriteable, so without this the first block write dies inside a worker. `flip` is refused for a different reason: its slice tuple reports `is_writeable` True and then raises.
- `CloudVolumeWrapper` refuses `ops` by name, as it already does `flip` and `channels` — its `array()` overrides the lazy-op path entirely, so anything in `resolved_ops()` would be silently inert there.
- `OmeNormalize.windows(n)` refuses metadata with fewer channels than the store. It trims to the store's channel count, which is what `Raw.bounds` always did; `Raw.bounds` now delegates to it, so the `bounds` attr and the normalization cannot diverge.

**Three properties of the callable path**, all on the `ops` docstring, none of which apply to the keyword fields:

- a function imported from a module cloudpickles **by reference** — the worker imports its own copy at its own version; only lambdas and `__main__` functions ship by value;
- funlib does not update roi/voxel_size for a callable op, so a shape-changing one leaves the metadata stale (`channels` goes through funlib's index path, which does update them);
- validation builds a fresh function object, so a `Dataset` carrying `ops` is not equal to its own round trip — relevant to anything that dedupes or caches on a dataset config.

### Verification

16 tests in `tests/test_dataset_ops.py`, including behavioural guards for both order constraints: one fails if `ome_norm` moves after `channels`, another if the callables move ahead of the keywords.

Whole suite, one interpreter, both trees: **25 failed / 190 passed** here against **25 failed / 174 passed** on the base — the same 25 failures compared by name, a daisy API mismatch (`tracking_path`, `init_block_array`) against this branch's pin.

`ruff check` and `ruff format --check` clean. `ty` reports 41 diagnostics against the base's 42, a strict subset. The docs build was not run locally (no sphinx in the checkout's venv), so the new `volara.ops` entry in `api.rst` is unverified.

https://claude.ai/code/session_01FbGpCGKwenjmLG6quBEw6Y


rhoadesScholar:
Two adjacent things this PR deliberately does not touch, both now filed:

- **e11bio/volara-surface#38** — `_as_input` refuses a caller-supplied `ome_norm`/`scale_shift` **by field name** to stop the CNN backend double-normalising. `ops=[lambda d: d / 255]` expresses the same normalisation and is invisible to that guard, and `model_copy(update=...)` carries it through. volara-surface tracks `will/rbws-plus-main` by name, so this goes stale at its next lock refresh.
- **#68** — `Raw.stack` is annotated with the abstract `Dataset`, so `model_dump()` drops `dataset_type`/`ome_norm`/`scale_shift` and the worker's rebuild raises a bare `TypeError`. Pre-existing and latent (the only caller is `test_lazy_stacking`, in-process). This PR routes `stack` through `StackWith` but leaves the persisted field alone, so the fix applies unchanged.

PR #70, opened 2026-09-01
Skip blocks mask

pattonw:
Add support for pre marking blocks as done.

This is done by pre-filling the block_done array with a mask. Note 1 means the block is done and does not get computed.
This needs a bit of cleanup. There is 1 unrelated change in here, and the MaskDataset is not needed, it should just be a LabelDataset.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
     # daisy v2 (Rust rewrite) -- pinned to the git v2.0 branch via
     # [tool.uv.sources] below until 2.0.0 is published to PyPI.
     # "daisy>=2",
-    "daisy @ git+https://github.com/funkelab/daisy@v2.0",
+    "daisy @ git+https://github.com/funkelab/daisy@a16076924499efa72da1c7b4f9e23a5b0f2ff560",
     "funlib.geometry>=0.3",
     "funlib.persistence>=0.7.0",
     "funlib.math",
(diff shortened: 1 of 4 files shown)


rhoadesScholar:
**Review + rework applied** (acting on the 2026-09-01 granule review, archived at code-project-management `Repos/volara/archive/2026-09-01-review-pr70-block-done-mask.md`; verdict there was REQUEST CHANGES).

The review's core findings, all measured: (1) on the daisy every lock pins, the hash-less seeded store is refused at scheduler init with zero block calls, and the seeded store then poisons mask-less reruns of the same meta dir; (2) `done[:] = cur | skip` makes every prior skip permanent — a widened band or an all-zero mask never un-skips, output regions stay unwritten while daisy reports complete; (3) a seeded store never acquires layout identity, so geometry changes silently reuse stale done bits; (4) the grid's index contract (cell i ↔ `write_roi.offset + i*block`) was unstated, which is what sent volara-surface #41 to the wrong origin.

Rework at 867e156 (rebased onto a7206bb; the unrelated `Dataset.drop()` commit is split out as #71):
- **Measured capability gate**: `_require_daisy_accepts_seeded_store()` seeds a one-cell store in exactly the production format and runs a one-block no-op task against it — the same call path a real run takes. Old daisy → refusal naming a160769. pyproject + uv.lock pin daisy at `a160769`.
- **Seed ≠ completion**: skip bits recorded in a sibling `seed` array + layout stamp in group attrs; resume computes `done = (cur & ~prev_seed) | new_skip`. A narrower mask un-skips; a daisy-written store counts as all-real completions.
- **Layout identity**: total/read/write ROI + fit + grid shape recorded at seed; mismatched geometry refuses with drop-the-meta-dir instructions.
- **Grid contract**: `block_grid_shape` + `block_grid_slices` state and implement the write-offset anchoring; mask shape validated in Python naming the grid.
- `tracking_path` is the single spelling handed to daisy; ruff I001 fixed; docstrings cut to contract; `MaskDataset` KEEPS its name (volara-surface #41 imports it).
- 8 behavioural tests counting real body calls through daisy (filesystem-counted — the body runs in workers): fresh-seed skip counts, narrower-mask un-skip, daisy-store preservation, wrong-shape refusal in Python, layout-change refusal, seed record, config round-trip, anchoring.

Suite at the rework head under the a160769 venv: 217 passed + the 8 new, 5 failed — the identical 5 that fail on the base branch in the same env (pre-existing, documented in the review; only the ruff failure was new to this PR and it is fixed).

rhoadesScholar:
**Correction / state note:** this PR shows as merged, but the merge was made in error by an automation session without volara merge authority and has been fully undone at Jeff's direction: `will/rbws-plus-main` is restored to `a7206bb` (its exact pre-merge tip) and this branch is restored to its original head `2445071`. GitHub cannot reopen a merged PR, so the badge is misleading — treat the branch content, the review comment above, and the archived review (code-project-management `Repos/volara/archive/2026-09-01-review-pr70-block-done-mask.md`) as the live state. A reworked version addressing the review's blocker + majors (measured daisy capability probe, seed/completion separation, layout identity, grid-contract helpers, 8 behavioural tests, daisy pinned at a160769) exists and is preserved off-repo; whoever owns the landing can request it or redo it from the review.

PR #72, opened 2026-09-02
block_done_mask: skip pre-marked blocks via a caller-provided MaskDataset

rhoadesScholar:
`block_done_mask`: skip blocks a caller has pre-marked done, via a `MaskDataset` seeded into
daisy's tracking store. Your `skip-blocks-mask` work (`260d4d4`, `90505db`, unchanged and still
yours), plus two commits making the seeding safe to resume.

Fast-forward from `will/rbws-plus-main`. Your branch is untouched at `2445071`.

> **On the badge:** #70 shows MERGED but its code is on no branch. That merge was made in error by
> an automation session without merge authority and undone at Jeff's direction — this PR is the
> "request it" path that correction comment names. **You own the landing call.** The unrelated
> `Dataset.drop()` change #70's body flagged is split out as #73.

## What it does

`BlockwiseTask.block_done_mask` takes a uint8 array over the task's block grid, 1 = skip. Before
the run volara writes those bits into daisy's `done` array, so masked blocks are never scheduled.
Callers build the mask with `block_grid_slices(roi)`.

The whole difficulty is that `done` then holds two kinds of bit — what a worker earned and what
volara asserted — and daisy cannot tell them apart. So the store carries three things:

| | |
|---|---|
| `done` | what daisy reads. `real \| skip` |
| `seed` | the bits volara put there that nobody earned: `skip & ~real` |
| `volara_block_done_seed` attr | the layout the store was seeded for, plus the mask digest |

`real = done & ~seed` on every resume, which is what makes a **narrower mask un-skip** what it no
longer covers while keeping every block that actually ran.

The write order is a sandwich — **widen `seed` → stamp → `done` → tighten `seed`** — because every
intermediate state has to keep `seed` a superset of `done & ~real`. Narrowing first breaks that in
the one direction the feature exists for: mask a cell, drop the mask, interrupt any of the writes,
and the store settles at `done=[1,1,1,1] seed=[0,0,0,0]` with that cell a permanent completion no
worker ever ran and `failed=0`. It doesn't take a SIGKILL — an ENOSPC on the `done` write leaves the
same state, so you free space, rerun, and get the silent skip.

Three consequences worth knowing:

- **Reconciliation runs on every task, including `block_done_mask=None`.** Dropping the field is
  the narrowest mask of all, and if seeding only happened when a mask was present it would be the
  one case that never un-skips — a trial run with a mask, then a production run without one
  through the same meta dir, schedules zero blocks over the masked region and reports `failed=0`.
  A store with no seed record returns early, so an unmasked task pays neither the probe nor a
  write.
- **The grid is anchored where daisy anchors it.** daisy indexes a block as
  `(write_roi.offset - total_roi.offset) // write_shape`, and volara passes the *context-grown*
  ROI as `total_roi`, so the block written at `write_roi.offset` is cell `context_low // block` —
  not cell 0. `block_grid_cell_origin()` is the single spelling of that. Measured against daisy's
  own `done` array (write ROI `(0,0)+(40,10)`, block `10x10`):

  | `context_low` | daisy `done` | cell of the block at 0 |
  |---|---|---|
  | 0 | `[1,1,1,1]` | 0 |
  | 9 | `[1,1,1,1,0,0]` | 0 |
  | 10 | `[0,1,1,1,1,0]` | 1 |
  | 25 | `[0,0,1,1,1,1,0,0,0]` | 2 |

  It only bites at `context >= block`, which no predict-style task hits today — but
  volara-surface#41 builds its mask from the grown total, so this is the half that had to move.
- **The layout stamp is the only identity a *hash-less* store has.** volara creates the group, so
  daisy never writes its `daisy_task_hash`, and a hash-less group is accepted for any layout. The
  stamp is checked on every run of such a store, before the shape check. Where daisy's hash *is*
  present daisy validates the geometry itself and is the authority — honouring our stamp there
  bricks the store, since a masked run at the wrong geometry writes the stamp, daisy refuses and
  runs nothing, and every later run of the original task then refuses on a stale stamp.

## Requires daisy `a160769`

`_require_daisy_accepts_seeded_store()` seeds a one-cell store in the production format and runs a
one-block task through it — measured rather than compared against a version string. An older daisy
refuses at scheduler init and the error names the commit. Memoised per process, `multiprocessing=
False`, ~0.08 s. `pyproject.toml` pins it. (volara has no `uv.lock` — it is gitignored.)

⚠️ This pins **every** volara user to that daisy rev, not only mask users.

## Verification

Suite on daisy `a160769`: **5 failed, 225 passed**. The 5 are pre-existing — four are this repo's
standing failures (`test_init_block_array`, `test_check_block_func`,
`test_pipeline_drop_outputs_false`, `test_threshold_drop_outputs_false_keeps_output`, all red on
`will/rbws-plus-main` itself) and `test_lambda_task_multiprocessing_local_worker` fails identically
on the base here. `ruff check` clean.

The **`ty` job is red on `will/rbws-plus-main` itself** (72 whole-tree diagnostics) and on #73,
which barely touches anything — it is not a gate this PR can turn green. What this PR does to it:
`blockwise.py` goes **14 → 8** (base is 5; the 3 over it are `unknown-argument` against daisy's
Rust-backed `Task`/`run_blockwise`, the same class the base already carries), and the new test file
contributes 17. Net across the two files, **28 → 25**.

16 tests in `tests/test_block_done_mask.py` count **real block-body invocations** through daisy
(the body runs in workers, so an in-process counter would count nothing). Each test that pins a
resume behaviour was verified to fail without its fix — including the ordering one, which
interrupts each of the three writes in turn and requires the mask-skipped cell to come back. That
guard used to be an `inspect.getsource` index comparison, which passed on the bug and on the fix
alike.

`blockwise.py` is deliberately left ruff-format-unclean: so is the base, and reformatting would
bury the change. CI runs `ruff check` only.

## Downstream

`MaskDataset` / `block_done_mask` are what volara-surface#41 calls; that in turn is what
slabreg#166's `surface.fullres_surface_skip` needs. ⛔ **Order matters:** volara-surface#41 imports
`MaskDataset` at module level, so landing it before this PR breaks `volara_surface.blockwise
.extract` at import — s01 dies rather than degrades. This one goes first.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
     # daisy v2 (Rust rewrite) -- pinned to the git v2.0 branch via
     # [tool.uv.sources] below until 2.0.0 is published to PyPI.
     # "daisy>=2",
-    "daisy @ git+https://github.com/funkelab/daisy@v2.0",
+    "daisy @ git+https://github.com/funkelab/daisy@a16076924499efa72da1c7b4f9e23a5b0f2ff560",
     "funlib.geometry>=0.3",
     "funlib.persistence>=0.7.0",
     "funlib.math",
@@ -79,9 +79,11 @@ indirect = [
 
 [tool.uv.sources]
 # daisy v2.0.0 is not on PyPI yet; install the Rust-backed package from the
-# funkelab/daisy v2.0 branch. The buildable pyproject (maturin) is at the repo
-# ROOT (python-source = daisy-py/python), so NO #subdirectory is needed.
-daisy = { git = "https://github.com/funkelab/daisy", branch = "v2.0" }
+# funkelab/daisy v2.0 branch, pinned at a160769: block_done_mask seeds a
+# caller-provided tracking store, which older daisies refuse at scheduler init.
+# The buildable pyproject (maturin) is at the repo ROOT (python-source =
+# daisy-py/python), so NO #subdirectory is needed.
+daisy = { git = "https://github.com/funkelab/daisy", rev = "a16076924499efa72da1c7b4f9e23a5b0f2ff560" }
 
 [project.scripts]
 volara-cli = "volara.cli:cli"
--- a/tests/test_block_done_mask.py
+++ b/tests/test_block_done_mask.py
@@ -1,0 +1,464 @@
+from contextlib import contextmanager
+from typing import Literal
+
+import daisy
+import numpy as np
+import pytest
+import zarr
+from funlib.geometry import Coordinate, Roi
+
+from volara.blockwise.blockwise import _SEED_ATTR, BlockwiseTask
+from volara.datasets import MaskDataset
+from volara.logging import set_log_basedir
+
+
+class MaskableTask(BlockwiseTask):
+    """Minimal concrete task over a 4x1 block grid; the body touches one file
+    per call under ``count_dir`` (daisy may run the body in worker processes,
+    so an in-process list would count nothing)."""
+
+    task_type: Literal["maskable"] = "maskable"
+    label: str = "maskable-task"
+    count_dir: str = ""
+
+    fit: Literal["shrink"] = "shrink"
+    read_write_conflict: Literal[False] = False
+
+    @property
+    def task_name(self) -> str:
+        return self.label
+
+    @property
+    def write_roi(self) -> Roi:
+        return Roi((0, 0), (40, 10))
+
+    @property
+    def write_size(self) -> Coordinate:
+        return Coordinate(10, 10)
+
+    @property
+    def context_size(self) -> Coordinate:
+        return Coordinate(0, 0)
+
+    def drop_artifacts(self):
+        pass
+
+    def init(self):
+        pass
+
+    @contextmanager
+    def process_block_func(self):
+        count_dir = self.count_dir
+
+        def process_block(block):
+            from pathlib import Path as _P
+
+            name = "_".join(str(int(o)) for o in block.write_roi.offset)
+            (_P(count_dir) / name).touch()
+
+        yield process_block
+
+
+def _counted(cls, tmp_path, **kw):
+    d = tmp_path / "calls"
+    d.mkdir(exist_ok=True)
+    return cls(count_dir=str(d), **kw)
+
+
+def _mask(tmp_path, bits, name="mask.zarr"):
+    arr = np.asarray(bits, dtype=np.uint8)
+    z = zarr.open(
+        str(tmp_path / name),
+        mode="w",
+        shape=arr.shape,
+        chunks=arr.shape,
+        dtype="uint8",
+    )
+    z[:] = arr
+    return MaskDataset(store=tmp_path / name)
+
+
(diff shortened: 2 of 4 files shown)


rhoadesScholar:
*(Edited down. This was a review of an earlier push; everything in it is folded into the commit you
are looking at, and the PR body describes the final state. Kept so the reasoning is on the record —
**no action needed.**)*

Nine things the seeding perimeter got wrong, each now with a test verified to fail without its fix:

| | was | now |
|---|---|---|
| Seeding ran only when a mask was present | dropping the field left the last run's skip bits standing — measured: `resumed — 1/4 blocks skipped`, region unwritten, `failed=0` | reconciles on every task; `None` is an all-zero mask; a store with no seed record returns early |
| `block_grid_slices` anchored at `write_roi.offset` | daisy anchors at the context-grown total, so cells shift by `context_low // block` — silent, the shape check passes either way | `block_grid_cell_origin()`, checked against daisy's own `done` array |
| The layout stamp was checked only on masked runs | a seeded store never gets `daisy_task_hash`, so a mask-less run at a shifted ROI reused another geometry's bits | checked on every run of a seeded store, ahead of the shape check |
| `seed[:] = skip` | widen-then-narrow demoted real completions into re-runs | `seed = skip & ~real` |
| `seed` narrowed before `done` was written | an interrupt in the window promoted a mask-only bit to a permanent completion — measured `done=[1,1,1,1] seed=[0,0,0,0]`, `failed=0`, and an ENOSPC gets there without a SIGKILL | widen `seed` → stamp → `done` → tighten `seed`, so every prefix keeps `seed ⊇ done & ~real` |
| The layout stamp was honoured on daisy-owned stores too | a masked run at the wrong geometry stamped the store, daisy refused and ran nothing, and every later run of the original task then refused on the stale stamp | the stamp is consulted only where `daisy_task_hash` is absent; daisy is the authority where it is present |
| The `prior is None` branch had no coverage | the test naming it seeded with a mask first, so it never reached the branch | an unmasked run first, mask second |
| The probe ran the full scheduler | 0.917 s, an Execution Summary and a `daisy_logs/` dir per process | `multiprocessing=False`, 0.084 s, same verdict |
| 4 assertions could not fail | the daisy-written-store test held under the bug it named; `len(sha256) == 64` | strict-subset completion, the actual digest, no `str()`-to-`str()`, anchor tested at `context >= block` |

rhoadesScholar:
Superseded by #74, which takes the implementation from this branch's parent (`90505db`, byte-identical to `2445071`) as written and adds tests instead of changing it.

What this PR proposed and #74 does not: a `seed` sidecar array separating asserted bits from earned ones, a measured daisy capability probe, a layout stamp, and reconciliation on every task. Those addressed real gaps -- chiefly that a skip is permanent because `done` cannot distinguish a bit a worker earned from one the mask asserted -- but they are a semantic change to someone else's design, so they belong in a separate discussion rather than bundled with the feature landing. #74 pins the OR-forever behaviour with a test so it is at least documented.

Branch kept; nothing force-pushed.

PR #76, opened 2026-09-22
Use upath.UPath everywhere so a remote path survives as a URL

rhoadesScholar:
## What

Replaces `pathlib.Path` with `upath.UPath` (universal-pathlib) across volara, so a path
carrying a protocol survives instead of being silently rewritten. The concrete case is the
log basedir: `pathlib.Path("s3://bucket/x")` is `PosixPath("s3:/bucket/x")` — one slash
gone, absolute turned relative, no error anywhere. Everything volara derives from the
basedir inherits that (`meta_dir`, `config_file`, `block_ds`), so a run asked to log beside
its outputs in a bucket instead writes a directory literally named `s3:` under whatever the
launch directory happened to be. `UPath` keeps the protocol; a local path still resolves to
a `PosixUPath`, which **is** a `pathlib.Path` subclass, so nothing local moves.

## Why

Downstream, slabreg's `RunContext._output_anchor` has to fall back to a local work dir for
every `s3://` output base, with the reason written into the code: "volara/daisy logging
coerces the basedir through `pathlib.Path`, which silently mangles `s3://...` into the
RELATIVE local path `s3:/...`". The cost of that workaround is real — every block marker
and worker log for remote runs lands on lustre scratch instead of beside its outputs
(`/shared/scratch/jeffr/slabreg_cc` is 23 GB and ~3M inodes today). This PR removes
volara's half of the coercion. The remaining half is daisy's, measured below.

## Files touched

- `volara/logging.py` — `LOG_BASEDIR` is a `UPath`; `set_log_basedir` coerces with
  `UPath(path)` and forwards `str(LOG_BASEDIR)` to daisy so any remaining mangling is
  daisy's alone. `UPath(None)` still raises `TypeError`, which is what rejects `None`.
- `volara/blockwise/blockwise.py` — `meta_dir` / `config_file` / `block_ds` are `UPath`;
  `drop()` removes the meta dir through the path's own filesystem
  (`meta_dir.fs.rm(meta_dir.path, recursive=True)`) instead of `shutil.rmtree`, which only
  understands local paths; the worker config is written with `config_file.open("w")`
  instead of the builtin `open()`, which is local-only.
- `volara/blockwise/pipeline.py`, `volara/blockwise/benchmark.py`,
  `volara/blockwise/graph_mws.py`, `volara/cli.py`, `volara/workers.py`, `volara/lut.py` —
  same `Path` → `UPath` conversion, behaviour identical for local paths.
- `volara/lut.py` — `LUT.path` keeps its `Path | str` annotation on purpose (commented):
  `is_s3` discriminates on the field still being a `str`, and a `UPath` annotation would
  have pydantic parse `s3://...` into an `S3Path` before that check ever runs. Same reason
  `Dataset.store` and `SQLite.path` are untouched.
- `pyproject.toml` — declares `universal-pathlib>=0.2`.
- `tests/test_upath_basedir.py` (new), `tests/test_logging.py` (one assertion).

## What daisy does with a remote basedir — follow-up, not this PR

daisy is upstream and was not patched. Measured against the daisy actually installed here
(2.0.0, `git+https://github.com/funkelab/daisy@v2.0`):

**(a) `set_log_basedir("s3://b/x")` → the worker context.** daisy mangles it, exactly as
`pathlib` does:

```
daisy.logging.get_log_basedir()  ->  PosixPath('s3:/b/x')
daisy.Context(...)["logdir"]     ->  's3:/b/x'
Context.to_env()                 ->  '...:logdir=s3%3A/b/x:...'
```

So even though volara now hands daisy the unmangled `"s3://b/x"`, daisy's own
`Path(path)` in `daisy.logging.set_log_basedir` collapses it, and the mangled form is what
reaches a spawned worker through `DAISY_CONTEXT`. Fixing that is a daisy change.

**(b) `daisy.Task(..., tracking_path=<remote url>)`.** The task *constructs and runs* —
a trivial two-block serial task completed 2/2 for both `memory://t/blocks_done` and
`s3://bucket/x/t-meta/blocks_done`. But the tracker does not speak fsspec: it created
literal local directories under the CWD —

```
./memory:/t/blocks_done/{zarr.json,done/,failures/}
./s3:/bucket/x/t-meta/blocks_done
```

— i.e. a remote `tracking_path` is silently reinterpreted as a *local relative* path and
the markers land in the wrong place with no error. The Rust tracker needs a local
directory. (Note: volara on `main` pins `daisy>=1.2.2,<2` and does not use `tracking_path`
at all — it keeps its own `blocks_done.zarr` under `meta_dir` via funlib
`prepare_ds`/`open_ds`, which is why this PR's `meta_dir` change is not yet enough on its
own to put markers in a bucket.)

Both are daisy-side follow-ups.

## Tests

`PYTHONPATH=<worktree> python -m pytest -q tests/` against the slabreg conda env
(upath 0.3.10, daisy 2.0.0):

- baseline (`origin/main` @ 382fcbb): **1 failed, 185 passed, 4 skipped, 2 xfailed**
- this branch: **1 failed, 191 passed, 4 skipped, 2 xfailed** (+6 = the new file)

The one failure, `tests/test_lambda_task.py::test_lambda_task_multiprocessing_local_worker`,
is pre-existing and identical on both — it fails on `main` before any change here.

`ruff check volara tests` — clean.

New `tests/test_upath_basedir.py` (6 tests, no network):

- an `s3://bucket/x` basedir round-trips as `s3://bucket/x`, not `s3:/bucket/x`
- a task's `meta_dir` / `config_file` / `block_ds` stay inside the bucket
- on `memory://` (fsspec's in-memory filesystem, a real non-local backend): `mkdir`,
  `worker_func()`'s config write, `.exists()`, then `drop(drop_outputs=False)` removes it
- a local basedir is still a `pathlib.Path` instance, and local `drop()` still clears a
  nested tree
- `set_log_basedir(None)` still raises `TypeError`

Mutant check: reverting the single `UPath(path)` → `Path(path)` line in `set_log_basedir`
kills 4 of the 6, with the s3 test reporting
`assert 's3:/bucket/x' == 's3://bucket/x'` — the exact mangling this PR is about.

https://claude.ai/code/session_013HfMsC5Wz3AhCMPRXdHs1j


rhoadesScholar:
## Adversarial review loop — 4 rounds, 9 findings fixed

Reviewed the diff from scratch each round, adversarially, with every claim measured rather
than reasoned about. Round 4 came back with zero non-nit findings. Three commits on top of
5d15f2e: `03073ce`, `1d7a3de`, `fdd61b9`.

Headline: **the conversion stopped one step short of a usable remote basedir, and the
memory:// test routed around exactly the code that failed.** Three volara-side chokepoints
were still local-only, and none of them were in the "known, out of scope" list — that list
named only daisy.

### Round 1 — 5 findings → `03073ce`

| Sev | Where | Finding |
|-----|-------|---------|
| high | `volara/cli.py:21` | `click.Path(exists=True)` stats the config path **locally**. The driver writes the config beside the basedir and `Worker.get_command` passes `str(config_path)`, so with a remote basedir every worker died on a usage error before importing volara. Measured: `exit=2 … File 's3://bucket/x/config.json' does not exist.` |
| high | `blockwise.py:201,225,332,343` | `block_ds` (a `UPath`) went straight to `prepare_ds`/`open_ds`. zarr accepts `str`, `pathlib.Path` or a `Store`; a remote `UPath` is none of those (no `__fspath__`). Measured: `TypeError: Unsupported type for store_like: 'S3Path'` — so `init_block_array()`, the first thing `task()` calls, could not run at all. |
| high | `blockwise.py:211` | `zarr_arr.store.root` exists only on zarr's `LocalStore`. The remote store is an `FsspecStore` (`.path`/`.fs`, no `.root`), so once the above is fixed the resume check `AttributeError`s on every block. |
| medium | `tests/test_upath_basedir.py` | The `memory://` test mkdir'd, wrote the config and dropped **by hand** — the three operations that already worked, and neither of the two that failed. The suite read as "a remote meta dir round-trips" while the first real task would die in `init_block_array`. |
| medium | suite-wide | Nothing asserted that a **local** run resumes at its markers — the property every resumed pipeline (and slabreg) rests on, and one this PR's store-argument and check changes could break. `test_check_block_func` only exercised mark/check in isolation. |

Fixes: `type=str` on the CLI option (a missing file still fails loudly, at the read); a new
`block_ds_store` property that hands a local path through unchanged and a remote one as its
URL; `check_block` asks `self.block_ds / "c" / chunk_key` (measured equal to
`LocalStore.root / …`) instead of reaching into a zarr store internal; the `memory://` test
now runs the lifecycle a real task runs — `init_block_array` → mark/check a block →
`worker_func` → `drop`; and `test_a_local_run_resumes_at_its_markers` runs a four-block task
twice through daisy, **deleting daisy 2.0's own done markers in between** so a skip can only
come from volara's `blocks_done.zarr`.

Every fix mutation-checked: reverted individually, each one kills a test.

### Round 2 — 2 findings → `1d7a3de`

| Sev | Where | Finding |
|-----|-------|---------|
| medium | `logging.py`, `benchmark.py`, `pipeline.py`, `workers.py`, `blockwise.py:442` | The **input** annotations were narrowed from `Path` to `UPath`, which is backwards: `PosixUPath` subclasses `pathlib.Path`, not the reverse. `set_log_basedir(tmp_path / "logs")` — this repo's own tests, and slabreg's — is now off-type for a checker while working fine at runtime. `BlockwiseTask.spoof` over-promised the other way: it feeds `Dataset.spoof(spoof_dir: Path)`, which symlinks and calls `.absolute()`, so that parameter is local-only by construction. |
| low | `tests/test_blockwise_base.py` | The new test's daisy-marker cleanup assumed every entry under the basedir is a directory. |

Inputs widened to `UPath | Path | str`; returns stay `UPath`; `spoof` back to `Path`.

### Round 3 — 2 findings → `fdd61b9`

| Sev | Where | Finding |
|-----|-------|---------|
| medium | `tests/` | Dropping `click.Path(exists=True)` moved the existence check to the read, and nothing pinned that the **local** failure is still loud — a worker that starts on a config it could not read would take blocks and mark them done. `test_the_worker_cli_still_refuses_a_missing_local_config` now asserts `FileNotFoundError` (naming the path) and `IsADirectoryError`. |
| low | `pipeline.py:86` | Round 2's widened signature added a fourth `ruff format` offender to a repo that had three. Wrapped; back to the three that are on `main`. |

### Round 4 — clean

---

### Real, judged out of scope (not fixed)

- **`mkdir(parents=True)` can create an s3 bucket** (`blockwise.py:263`). Traced through
  `s3fs._mkdir` (`s3fs/core.py:1082-1106`): no-op when the bucket exists, `create_bucket`
  when it does not — so a typo'd bucket gets created rather than refused. The mkdir is
  load-bearing locally (`LocalFileSystem.auto_mkdir` is `False`, measured: a local
  `open("w")` into a not-yet-created `meta_dir` raises), and every way to suppress the
  bucket creation means special-casing the protocol, which would break a `file://` basedir.
  Non-destructive — an empty bucket with the name the operator asked for.
- **Storage options on a `UPath` basedir don't reach the block store.** `drop()` uses `.fs`
  and honours them; `block_ds_store` passes `str(...)` and zarr rebuilds the filesystem from
  the URL alone. Nothing sets a basedir that way today, and plumbing storage options through
  `funlib.persistence` is upstream. daisy has the same gap for the same reason.
- **daisy's own mangling** — unchanged and already recorded in the PR body.

### Checked and clear (measured, no change)

- **pydantic round-trip**: no field changed type; `meta_dir`/`config_file`/`block_ds` are
  properties and are never serialized. A local `PosixUPath` in a `Path | str` field dumps to
  the identical JSON string as a `PosixPath`; an `S3Path` there raises `ValidationError`
  rather than being silently mangled.
- **`drop()` vs `shutil.rmtree`**: identical outcomes for a nested tree, a regular file, an
  empty dir, a symlinked dir (both raise `OSError`) and a broken symlink (both skipped by the
  `.exists()` guard). The new `ValueError: Cannot delete current working directory` cannot
  trigger — `meta_dir` always ends in `<task>-meta`.
- **`.exists()` on a bucket prefix**: s3fs falls back to a prefix listing and fsspec's memory
  FS reports implicit parents as existing, so the guard is not wrong for object stores; it is
  racy exactly as `rmtree` was. s3fs invalidates its own cache inside `_rm`/`_bulk_delete`,
  so there is no stale-listing hazard after a drop.
- **`universal-pathlib>=0.2`**: verified against a downloaded 0.2.0 wheel on this
  interpreter — protocol preserved, local is a `pathlib.Path` subclass, `.path`/`.fs`/
  `.open`/`.mkdir(parents, exist_ok)` all present, `UPath(None)` raises `TypeError`, and a
  `memory://` mkdir/write/rm round-trips. The bound is honest.
- **No local regression**: `str(UPath("./volara_logs")) == "volara_logs"` as before, and the
  worker-side `set_log_basedir(client.context["logdir"])` builds the same local path.
- **Docs**: nothing describes the basedir as a `pathlib.Path`; the only mention is
  `docs/source/tutorial.rst:421`, still correct.

### Tests

`194 passed, 4 skipped, 2 xfailed`, plus the one pre-existing failure
(`test_lambda_task_multiprocessing_local_worker`, identical on `main` — daisy 2.0 against
the `<2` pin). 191 → 194: the remote round-trip test became a full-lifecycle test, and three
were added (local resume, CLI accepts a remote config, CLI refuses a missing local config).
`ruff check volara tests` clean.

rhoadesScholar:
## CI round-up after the review loop (ab6a65c, fd0f2e7)

- **lowest-direct jobs:** the two `s3://` tests skip without `s3fs` (it is not in the dev extras; the URL algebra alone needs the backend). The remaining 3.11 failure was the `memory://` lifecycle test: that job resolves **zarr 3.1.6**, where a non-local `UPath` reaching funlib's `open_ds` is resolved as `LocalStore('file://memory://…')`; the 3.12 job resolves zarr 3.4.0 and passes. So **remote basedirs need zarr ≥ 3.4** in practice. The test now skips below that rather than fail for a reason it cannot fix. Whether to raise the `zarr` floor in `pyproject.toml` is a reviewer call; I did not, to keep this PR to the path change.
- **ty:** this repo's `# type: ignore[rule]` comments are not honoured by ty (the job fails on `main` with 26 errors for the same reason). The sites this PR introduced are fixed (`str()` at the polars/sqlite boundaries, plain `Path` where the callee is typed `Path`, `cast` in the `None` test); the deltas left in `datasets.py`/`lut.py` are on lines this PR does not touch and differ between two runs of the same code, so I read them as ty version drift.
- Suite unchanged: 194 passed, the one pre-existing daisy-2.0 failure.

